PR #6, opened 2018-06-05
Fix legacy code tooling details

peterblazejewicz:
Not sure if the legacy tests are still run, but here is a fix for nested source tree of assets being copied by `copy` module. The `cpy-cli` is by @sindresorhus:

- fix relative path
- swap `copy` with `cpy-cli` to use flattened structure of assets instead of
copying entire source tree: see @jonschlinkert/copy#23

I have no Windows machine setup at hand at the moment so I cannot verify if the tests actually fully succeed. But they at least run.

Thanks!


christopheranderson:
This is a good suggestion, but I might also be removing all the legacy tests. Cool if we wait until next week on this change?

peterblazejewicz:
> Cool if we wait until next week on this change?

Yup, no prob here. I've just read your comments on the DocumentDB repo yesterday.

christopheranderson:
removed the legacy tests (in dev), so this isn't needed anymore

PR #8, opened 2018-06-12
Port Version 1.14.4 release

christopheranderson:
This ports the changes from the v1.14.4 release of document-node (https://github.com/Azure/azure-documentdb-node/commit/ca337c9ddff136e01d6f589a45e24957d3efe5e3?diff=split&w=1) into the azure-cosmos-js library.

See the changelog.md file for all the relevant changes.

There are changes that are included that resolve known issues with PRs open against them in the documentdb-node repo

### DIFF
--- a/samples/DocumentManagement/app.js
+++ b/samples/DocumentManagement/app.js
@@ -44,6 +44,8 @@ var client = new DocumentDBClient( host, { masterKey: masterKey });
 // 5.1                  - Update some properties and replace the document
 // 5.2                  - Use ETag and AccessCondition to only replace document if it has not changed
 // 6. upsertDocument    - Update a document if it exists, else create new document
+// 6.1                  - Insert a document which does not exist
+// 6.2                  - Update a document which exists
 // 7. deleteDocument    - Given a document id, delete it
 //-------------------------------------------------------------------------------------------------------
 
@@ -165,17 +167,40 @@ init(function (err) {
                                                         }
 
                                                         //6.
-                                                        console.log('\n6. deleteDocument \'' + docLink + '\'');
-                                                        client.deleteDocument(docLink, function (err) {
+                                                        doc.id = "AndersenFamily2.0"
+                                                        doc.foo = "bar"
+                                                        doc.isUpdated = false
+                                                        var upsertDocLink = collLink + "/" + doc.id
+                                                        client.upsertDocument(collLink, doc, {}, function (err, insertedDoc) {
                                                             if (err) {
                                                                 handleError(err);
                                                             } else {
-                                                                console.log('Document deleted');
-                                                                
-                                                                //cleanup & end
-                                                                console.log('\nCleaning up ...');
-                                                                finish();
+                                                                console.log("\n6.1 document upsert when a document with the same id does not exist");
+                                                                console.log('The \'' + insertedDoc.id + '\' family has isUpdated set to \'' + insertedDoc.isUpdated + '\'');
                                                             }
+                                                            insertedDoc.isUpdated = true;
+                                                            client.upsertDocument(collLink, insertedDoc, {}, function (err, updatedDoc) {
+                                                                if (err) {
+                                                                    handleError(err);
+                                                                } else {
+                                                                    console.log("\n6.2 document upsert when a document with the same id exists");
+                                                                    console.log('The \'' + updatedDoc.id + '\' family has isUpdated set to \'' + updatedDoc.isUpdated + '\'');
+                                                                }
+
+                                                                //7.
+                                                                console.log('\n7. deleteDocument \'' + docLink + '\'');
+                                                                client.deleteDocument(docLink, function (err) {
+                                                                    if (err) {
+                                                                        handleError(err);
+                                                                    } else {
+                                                                        console.log('Document deleted');
+
+                                                                        //cleanup & end
+                                                                        console.log('\nCleaning up ...');
+                                                                        finish();
+                                                                    }
+                                                                });
+                                                            });
                                                         });
                                                     });
                                                 });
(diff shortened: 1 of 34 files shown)


samples/DocumentManagement/app.js
```diff
@@ -44,6 +44,8 @@ var client = new DocumentDBClient( host, { masterKey: masterKey });
 // 5.1                  - Update some properties and replace the document
 // 5.2                  - Use ETag and AccessCondition to only replace document if it has not changed
 // 6. upsertDocument    - Update a document if it exists, else create new document
+// 6.1                  - Insert a document which does not exist
+// 6.2                  - Update a document which exists
```
srinathnarayanan: As discussed offline, let us write some of these samples in typescript as well.

christopheranderson: We'll do this in the object-model-refactoring updates.

---
In reply to: [196175664](https://github.com/Azure/azure-cosmos-js/pull/8#discussion_r196175664) [](ancestors = 196175664)

src/queryMetrics/clientSideMetrics.ts
```diff
@@ -0,0 +1,58 @@
+import { QueryMetricsUtils } from "./queryMetricsUtils";
+
+export class ClientSideMetrics {
+    private requestCharge: number;
+    constructor(requestCharge: number) {
+        // Constructor
+        if (!QueryMetricsUtils.isNumeric(requestCharge)) {
```
bchong95: Is this check needed still?

src/queryMetrics/clientSideMetrics.ts
```diff
@@ -0,0 +1,58 @@
+import { QueryMetricsUtils } from "./queryMetricsUtils";
+
+export class ClientSideMetrics {
+    private requestCharge: number;
```
bchong95: Can we make this readonly now that we have the keyword?

src/queryMetrics/clientSideMetrics.ts
```diff
@@ -0,0 +1,58 @@
...
+    constructor(requestCharge: number) {
+        // Constructor
+        if (!QueryMetricsUtils.isNumeric(requestCharge)) {
+            throw new Error("requestCharge is not a numeric type");
+        }
+
+        this.requestCharge = requestCharge;
+    }
+
+    /**
```
bchong95: Do we need these annotation comments still?

src/queryMetrics/clientSideMetrics.ts
```diff
@@ -0,0 +1,58 @@
...
+     * Gets the RequestCharge
+     * @memberof ClientSideMetrics
+     * @instance
+     * @ignore
+     */
+    public getRequestCharge() {
```
bchong95: Can we use the getters / setters that supported by typescript?

src/queryMetrics/clientSideMetrics.ts
```diff
@@ -0,0 +1,58 @@
...
+     * Adds one or more ClientSideMetrics to a copy of this instance and returns the result.
+     * @memberof ClientSideMetrics
+     * @instance
+     */
+    public add(clientSideMetricsArray: ClientSideMetrics[]) {
+        if (arguments == null || arguments.length === 0) {
+            throw new Error("arguments was null or empty");
+        }
+
+        clientSideMetricsArray.push(this);
```
bchong95: push to a copy of clientSideMetricsArray so we don't modify the input.

src/queryMetrics/clientSideMetrics.ts
```diff
@@ -0,0 +1,58 @@
...
+    public getRequestCharge() {
+        return this.requestCharge;
+    }
+
+    /**
+     * Adds one or more ClientSideMetrics to a copy of this instance and returns the result.
+     * @memberof ClientSideMetrics
+     * @instance
+     */
+    public add(clientSideMetricsArray: ClientSideMetrics[]) {
```
bchong95: Can we use this:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters

since the original intent was for a variable length number of args.

src/queryMetrics/queryMetrics.ts
```diff
@@ -0,0 +1,369 @@
...
+     * @instance
+     * @ignore
+     */
+    public toDelimitedString() {
+        return QueryMetricsConstants.RetrievedDocumentCount + "=" + this.retrievedDocumentCount + ";"
+            + QueryMetricsConstants.RetrievedDocumentSize + "=" + this.retrievedDocumentSize + ";"
+            + QueryMetricsConstants.OutputDocumentCount + "=" + this.outputDocumentCount + ";"
+            + QueryMetricsConstants.OutputDocumentSize + "=" + this.outputDocumentSize + ";"
+            + QueryMetricsConstants.IndexHitRatio + "=" + this.getIndexHitRatio() + ";"
+            + QueryMetricsConstants.TotalQueryExecutionTimeInMs +
```
bchong95: If possible can we keep this in one line, since the format is key = value. If the line is too long, then we can leave it as is.

src/queryMetrics/clientSideMetrics.ts
christopheranderson: If we want to error on the client side, yes. TypeScript is a build time only type system, so it's possible to ignore it. Is this something we should aggressively guard against?

christopheranderson: Removed since this isn't meant to be used by customers, so we don't need to aggressively typecheck.

src/queryMetrics/clientSideMetrics.ts
christopheranderson: Removed.

src/queryMetrics/clientSideMetrics.ts
christopheranderson: In this case, if we make it readonly, we can just make it public, right?

src/queryMetrics/clientSideMetrics.ts
christopheranderson: That's wasteful. Instead, I'll initialize requestCharge to this.requestCharge.

PR #11, opened 2018-06-13
Fix incorrect paths in legacy test files

southpolesteve:
I saw in https://github.com/Azure/azure-cosmos-js/pull/6 that these tests might get removed? In the short term, they are preventing me from running the rest of the tests suite since require is throwing

### DIFF
--- a/src/test/legacy/aggregateQueryTests.js
+++ b/src/test/legacy/aggregateQueryTests.js
@@ -23,7 +23,7 @@
 
 "use strict";
 
-var lib = require("../..//"),
+var lib = require("../../../../../"),
     assert = require("assert"),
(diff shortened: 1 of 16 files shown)


southpolesteve:
Gah. Looks like my editor clobbered a file with prettier changes. Will update shortly

christopheranderson:
Those paths don't look right. The current paths have an extra slash at the end, which we should fix, but it's pointing to the right spot  otherwise. You're pointing really deep down the tree.

christopheranderson:
Or just remove "copy-legacy" from your "build" command for now. We're going to remove them.

christopheranderson:
removed the legacy tests (in dev), so this isn't needed anymore

PR #25, opened 2018-06-27
Port over existing docs to TS files

southpolesteve:
Moves the existing docs from `documentclient.ts`. There is still a lot of doc work to be done, the goal here was just to port current docs.

Notes:
- Marks everything in `documentclient.ts` as hidden
- Removes existing docs from `documentclient.ts`. We may not want to do this, the purpose was to help me keep track of progress.

### DIFF
--- a/src/documents/IndexingPolicy.ts
+++ b/src/documents/IndexingPolicy.ts
@@ -2,9 +2,12 @@ import { IndexingMode } from ".";
 import { DataType, IndexKind } from "./documents";
 
 export interface IndexingPolicy {
+    /** The indexing mode (consistent or lazy) {@link IndexingMode}. */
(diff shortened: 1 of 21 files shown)


src/documents/IndexingPolicy.ts
```diff
@@ -2,9 +2,12 @@ import { IndexingMode } from ".";
 import { DataType, IndexKind } from "./documents";
 
 export interface IndexingPolicy {
+    /** The indexing mode (consistent or lazy) {@link IndexingMode}. */
```
southpolesteve: There are several places we previously had links that typedoc makes unnecessary. Here it will link to `IndexingMode` automatically.

christopheranderson:
lgtm :shipit:

PR #27, opened 2018-06-29
Fix tests timeouts against production

southpolesteve:
One TTL test is still failing as a timeout


src/test/common/setup.ts
```diff
@@ -0,0 +1,9 @@
+process.on("unhandledRejection", (error) => {
```
christopheranderson: I dig this

PR #28, opened 2018-06-30
Release/2.0.0-2

christopheranderson:
- Update version to 2.0.0-2 in package.json and constants.ts
- Update readme sample with new OM


christopheranderson:
Killing this one in favor of #36

PR #30, opened 2018-07-05
Enable Travis CI

southpolesteve:
Supersedes https://github.com/Azure/azure-cosmos-js/pull/29. It appears that Travis needs to see the `travis.yml` file on master before builds will start working correctly.


christopheranderson:
Given that this is an additive engineering change, I don't think we need further review on this, for now.

PR #31, opened 2018-07-05
WIP Enable noUnusedLocals for typescript compiler

southpolesteve:
This PR naively fixes all issues raised by enabling `noUnusedLocals` in the typescript compiler. Still, work to be done as it required removing several private properties that are likely being used or don't have test coverage


southpolesteve:
Too many conflicts to keep this PR going. Will revisit later

PR #33, opened 2018-07-09
Add Prettier

southpolesteve:
- Adds prettier with default behavior
- Adds script `npm run format` for use in dev
- Adds script `npm run check-format` for use in CI

Prettier has two defaults which I'm not opposed to changing: 
- 80 character line length
- Double quotes for strings


package.json
```diff
@@ -20,6 +20,8 @@
   },
   "scripts": {
     "clean": "rimraf lib",
+    "format": "prettier --write \"src/**/*.ts\"",
+    "check-format": "prettier --list-different \"src/**/*.ts\"",
```
christopheranderson: I think we should have the build fail if there is formatting issues (similar to lint failures). Could this get added to the build command?

southpolesteve: Yep. I'll add

southpolesteve:
There is no config right now. We can put it in `.prettierrc` or in `package.json#prettier`

christopheranderson:
Let's put the config in now so its easier to configure later.

christopheranderson:
Dealers choice on a separate file or package.json approach. I like both.

PR #38, opened 2018-07-09
Add Travis to dev branch

southpolesteve:


southpolesteve:
@christopheranderson This should be good now. I setup Travis to only run one test at a time to avoid conflicts. We can change that once I get everything using `TestHelpers`

src/test/integration/session.spec.ts
```diff
@@ -50,7 +50,6 @@ describe("Session Token", function () {
         }
     };
 
-    afterEach(async function () { await TestHelpers.removeAllDatabases(client); });
```
christopheranderson: If we move to TestHelpers, we'll need to change the logic anyway for this, so I think you can ship anyway, but it's worth pointing out that that isn't going to speed anything up but might leave some tests laying around.

southpolesteve: I can't explain it, but this test seems to be flakey against staging. I've observed it locally and removing this line made it pass reliably.

southpolesteve: Of course, after I merged this it failed. Locally this test failed 2 out of 7 times for me tonight. Tomorrow I'll refactor to use test helpers which should solve the issue.

PR #49, opened 2018-07-13
Refactor of TestHelpers and session.spec test to be less flaky

southpolesteve:
This should fix the flakiness of the session spec test. There are still other places were not using test helper yet, but they are also not flaky :)

I also refactored test helpers to be functions. It didn't need a class wrapper.


src/test/integration/session.spec.ts
```diff
@@ -201,31 +202,31 @@ describe("Session Token", function() {
...
-
-    await client2.database(databaseDef.id).containers.create(containerDefinition, containerOptions);
-
-    await client2
-      .database(databaseDef.id)
-      .container(containerDefinition.id)
-      .read();
-    assert.equal(client.documentClient.getSessionToken(container.url), ""); // TODO: _self
-    assert.notEqual(client2.documentClient.getSessionToken(container.url), "");
+    // const client2 = new CosmosClient({
```
southpolesteve: Commented out because this test is skipped

src/test/integration/session.spec.ts
```diff
@@ -235,8 +236,7 @@ describe("Session Token", function() {
       consistencyLevel: ConsistencyLevel.Session
     });
```
southpolesteve: This is the actually important part of the PR. It should make the session test not be flaky anymore

christopheranderson:
I'll wait to review until after the CI is passing.

southpolesteve:
Reran and it passed 😭This PR is for making the tests LESS flaky. I suspect I was running tests locally at the same time, which this PR does not address.

src/test/integration/session.spec.ts
christopheranderson: It looks like the only change here is using the test helper to get the database. If that's it, then yeah; makes sense.

PR #51, opened 2018-07-14
Feature/#40 (continued)

### LINKED ISSUE #40: Add reference object to Response
Problem:
`.get` is confusing, but we need a way to get a reference to a given resource. We decided on a combo approach of renaming .get and adding the reference to the response of async calls.
From #39: 

> 4. Return reference object in response from async operations.
    - Example: `const { resource: db} = client.databases.create({id: foo});`
    - Pros: Helps reduce number of lines of code for many scenarios (like db and container create)
    - Cons: Doesn't remove need for `.get` because we don't want to force an async operation just to get a reference. Won't work with Steve's idea for simplifying the response types.

Additionally, we'll rename `result` to `body` to 

Implementation details:

```typescript
interface CosmosResponse<T, U> {
    body: T;
    headers: Headers;
    ref: U;
}

interface DatabaseResponse extends CosmosResponse<DatabaseDefinition, Database> {
    body: DatabaseDefinition;
    headers: Headers;
    ref: Database;
    database: Database;
}

/* Similar pattern for other types */
```

christopheranderson:
This finishes out the rest of the types for issues #40.

Also updates the README to new OM and the ref types

fixes #40

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,4 @@
 {
-    "mocha.files.glob":"test/legacy/**/*.js"
+    "mocha.files.glob":"test/legacy/**/*.js",
+    "editor.formatOnSave": true
 }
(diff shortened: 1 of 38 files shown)


christopheranderson:
Switched to a different PR See #52

PR #55, opened 2018-07-17
Symbols for storing headers and body

southpolesteve:


southpolesteve:
Not going to merge right now. User studies showed it was a bit confusing and this branch had some type hacks to make it work. If we do this later we'll start over in a new branch.

PR #60, opened 2018-07-19
Update docs for preview release

christopheranderson:
Just opening this PR for visibility right now. Still have work to be done.

Link to staged version of the docs: https://cosmosdbjspreview.z22.web.core.windows.net/classes/container.html


christopheranderson:
Ready for review

src/CosmosClient.ts
```diff
@@ -25,14 +63,38 @@ export class CosmosClient {
...
 
+  /**
+   * Used for reading, updating, or deleting a existing database by id or accessing containers belonging to that database.
+   *
+   * This does not make a network call. Use `.read` to get info about the database after getting the {@link Database} object.
+   *
+   * @param id The id of the database.
+   * @example Create a new container off of an existing database
+   * ```typescript
+   * const container = client.database("<database id>").container("<container id>");
```
southpolesteve: This example text says "create" so I would change this to:

`client.database("<database id>").containers.create("<container id>")`

Or change the example text

src/client/Conflict/Conflict.ts
```diff
@@ -4,20 +4,41 @@ import { RequestOptions } from "../../request";
...
+  /**
+   * Returns a reference URL to the resource. Used for linking in Permissions.
+   */
   public get url() {
     return `/${this.container.url}/${Constants.Path.ConflictsPathSegment}/${this.id}`;
   }
   private client: CosmosClient;
+  /**
+   * @hidden
+   * @param container The parent {@link CosmosClient} for the Database Account.
```
southpolesteve: Should link to `Container` and not `CosmosClient`

src/client/Container/Container.ts
```diff
@@ -9,43 +9,122 @@ import { UserDefinedFunction, UserDefinedFunctions } from "../UserDefinedFunctio
...
 
+  /**
+   * Used to read, replace, or delete a specific, existing {@link Item} by id.
+   *
+   * Use `.items` for creating new items, or querying/reading all items.
+   *
+   * @param id The id of the {@link Item}.
+   * @param partitionKey The partition key of the {@link Item}. (Required for partitioned containers).
+   * @example Replace an item
+   * const {body: replacedItem} = await container.item("<item id>").replace({id: "<item id>", newProp: {}});
```
southpolesteve: `({id: "<item id>", newProp: {}})` is a bit weird. It makes me think that all updates have to be passed in via a sub-object. What about something more real world? `({id: "<item id>", title: "Updated post", authorID: 5 })`

src/client/Item/Item.ts
```diff
@@ -4,17 +4,50 @@ import { RequestOptions, Response } from "../../request";
...
     this.client = this.container.database.client;
   }
 
+  /**
+   * Read the item's definition.
+   *
+   * There is no set schema for JSON items. They may contain any number of custom properties.
+   *
+   * @param options Additional options for the request, such as the partition key.
+   * Note, if you provide a partition key here, it will override the primary key on this object.
```
southpolesteve: Can you be more specific on this line? Kinda sounds like a new primary key will be written?

src/client/Item/Item.ts
```diff
@@ -4,17 +4,50 @@ import { RequestOptions, Response } from "../../request";
...
+   */
   public read(options?: RequestOptions): Promise<ItemResponse<any>>;
+  /**
+   * Read the item's definition.
+   *
+   * Any provided type, T, is not necessarily enforced by the SDK.
```
southpolesteve: Do we provide an example of this anywhere? Not sure it needs to be inline, but could be linked from these methods.

src/client/Item/Items.ts
```diff
@@ -6,34 +6,109 @@ import { Container } from "../Container";
 import { Item } from "./Item";
 import { ItemResponse } from "./ItemResponse";
 
+/**
+ * Operations for creating new containers, and reading/querying all containers
```
southpolesteve: This entire block talks about containers but should be items.

src/documents/ConnectionMode.ts
```diff
@@ -0,0 +1,5 @@
+/** Determines the connection behavior of the CosmosClient. Note, we currently only support Gateway Mode. */
+export enum ConnectionMode {
+  /** Gateway mode talks to a intermediate gateway which handles the direct communicationi with your individual partitions. */
+  Gateway = 0
```
southpolesteve: Why is this a numeric enum? Typescripts many ways of doing enums confuse me.

src/documents/ConsistencyLevel.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * <p>Represents the consistency levels supported for Azure Cosmos DB client operations.<br>
```
southpolesteve: Can we link to the portal docs for this? I think they are pretty good.

src/CosmosClient.ts
christopheranderson: Ah, good catch.

src/documents/ConnectionMode.ts
christopheranderson: This is a numeric enum because that's what it was in v1 and we haven't done much implementation changes under the hood.

image

I'm open to changing the enum value, but we should do that in a separate PR that addresses the other implementation issues.

src/client/Item/Item.ts
christopheranderson: I provided 1 example, but I think when we have proper docs, we can have a link to a section on our generics support. Our support for generic is kinda flakey, since we really only support interfaces. Classes will "work", but if you "typeof" our response, it will always be "object".

christopheranderson: We might be able to add a feature where you pass a class which we'll new up along on the options object, but that's a separate issue.

southpolesteve: That would be cool!

PR #64, opened 2018-07-23
Refactor Tests to use Test Helpers

southpolesteve:
Only one test failing! Thanks typescript. This started out as trying to convert all the tests to use `getTestDatabase` but ended up doing many things:

- Remove all `// tslint:disable:max-line-length` comments. Prettier handles this automatically.
- Convert test config file to export { endpoint, masterKey } 
- Allow `getTestDatabase` and `removeAllDatabases` to use default client
- Swap argument order of `getTestDatabse` and `getTestContainer` so `client` is easier to omit
- Remove lots of unnecessary comments, client instantiations, and comments


src/test/integration/aggregateQuery.spec.ts
```diff
@@ -5,14 +5,11 @@ import { Container, ContainerDefinition, Database } from "../../client";
...
-import testConfig from "./../common/_testConfig";
+import { endpoint, masterKey } from "./../common/_testConfig";
 import { TestData } from "./../common/TestData";
 import { bulkInsertItems, getTestContainer, removeAllDatabases } from "./../common/TestHelpers";
 
 // process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0";
-
-const endpoint = testConfig.host;
-const masterKey = testConfig.masterKey;
 const client = new CosmosClient({ endpoint, auth: { masterKey } });
```
christopheranderson: Probably don't need a client here, either, right?

src/test/integration/documentClient.spec.ts
```diff
@@ -1,14 +1,12 @@
 ﻿import * as assert from "assert";
 import { Constants, CosmosClient, IHeaders } from "../../";
-import testConfig from "./../common/_testConfig";
+import { endpoint, masterKey } from "./../common/_testConfig";
+
+const client = new CosmosClient({ endpoint, auth: { masterKey } });
```
christopheranderson: Remove unnecessary client

southpolesteve: This one is still used here. But this test is also testing `client.documentClient` directly so we should probably just delete it entirely...

christopheranderson:
VSTS failed. Kicked off another build.

christopheranderson:
Build 12 succeeded, but VSTS doesn't seem to track that the status has changed: https://cosmos-db-sdk-public.visualstudio.com/cosmos-db-sdk-public/_build/results?buildId=12&view=results 

I'm treating it as a passing on both travis and VSTS, and merging.

src/test/integration/documentClient.spec.ts
christopheranderson: Ah yeah, documentClient leaking is still a hack we've done to expose the session management stuff until we refactor documentClient away.

PR #65, opened 2018-07-23
First crack at parallel tests

southpolesteve:
#64 paves the way for parallel tests. This is a first crack at getting that working. On my local machine running tests against a staging endpoint (in a region I am not currently close to) it takes the test run from 7m to 4m.


christopheranderson:
If we want tests to run in parallel, maybe we should make it the default so we catch breaks easier. Running sync is really a feature to tracking down issues, not the nominal case for us.

southpolesteve:
This needs a restart. Still think there is value here, but not gonna pursue it via this PR

PR #83, opened 2018-07-30
Add system generated keys for permissions and items

southpolesteve:
Adds system generated keys to the return bodies for `Permission` and `Item` calls

@christopheranderson any thoughts on the approach of using `<T & ItemBody>` for items? Open to other suggestions, this was the option that made sense to me in the TS docs but seems like `<ItemBody<T = any>>` could also work.

### DIFF
--- a/src/client/Item/Item.ts
+++ b/src/client/Item/Item.ts
@@ -2,6 +2,7 @@ import { UriFactory } from "../../common";
 import { CosmosClient } from "../../CosmosClient";
 import { RequestOptions, Response } from "../../request";
 import { Container } from "../Container";
+import { ItemBody } from "./ItemBody";
 import { ItemResponse } from "./ItemResponse";
 
 /**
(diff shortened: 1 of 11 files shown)


christopheranderson:
I sent a follow up PR with some suggested changes: https://github.com/Azure/azure-cosmos-js/pull/84

src/client/Item/Item.ts
```diff
@@ -68,7 +69,9 @@ export class Item {
     if ((!options || !options.partitionKey) && this.primaryKey) {
       options.partitionKey = this.primaryKey;
     }
-    const response = await (this.client.documentClient.readDocument(this.url, options) as Promise<Response<T>>);
+    const response = await (this.client.documentClient.readDocument(this.url, options) as Promise<
+      Response<T & ItemBody>
```
christopheranderson: Unless we also change the `<any>` signature, this won't be seen.

christopheranderson:
:shipit:

Issue #73, opened 2018-07-25
Support callback based forEach


christopheranderson:
We could support an overload where callbacks are supported for forEach, this would help with issues where certain versions of Node doesn't have the AsyncIterator symbol yet.

See #71

Issue #71, opened 2018-07-24
forEach's asyncIterator doesn't work out of the box below Node.js 10


christopheranderson:
Most folks who want to use forEach's AsyncIterator need to use a polyfill to make it work.

See https://github.com/Microsoft/TypeScript/issues/14151

Polyfill: 
```typescript
(<any>Symbol).asyncIterator = Symbol.asyncIterator || Symbol.for("Symbol.asyncIterator");
```

I'm going to propose that we add this polyfill to the typedocs to make it easier to figure out.

I'm going to open a second issue to track adding the callback based model back (simulatenous support)

PR #87, opened 2018-07-31
Update forEach/Adds AsyncIterator

christopheranderson:
Switches the forEach implementation to be AsyncIterator instead.

Additionally, it changes the forEach implementation to accept a callback

fixes https://github.com/Azure/azure-cosmos-js/issues/73

fixes https://github.com/Azure/azure-cosmos-js/issues/71


christopheranderson:
Travis is sick. Merging anyway since VSTS is fine.

PR #103, opened 2018-08-13
Add script for analyzing webpack bundle

southpolesteve:


christopheranderson:
Is this useful still, @southpolesteve, now that we're moving to rollup?

southpolesteve:
Gonna take a different approach. https://bundlephobia.com/result?p=@azure/cosmos@2.1.1 is good enough for now to check size and contents. I also think some kind of CI check on bundlesize would be good.

PR #104, opened 2018-08-13
ES module interop fixes; use rollup for production builds

bterlson:
This pull request replaces Webpack (sorry Sean 😋) with Rollup and produces a single file entry point under `dist` with the library's entire contents flattened out. This is often called a "flattened ES module".

Distributing flattened commonjs modules has a significant impact on the load times. On my machine, I see the following results for time-to-require cosmos-js:

* Before PR: ~280ms
* After PR: ~25ms

In order to implement this PR, I had to make a number of changes:

* Use standard es6 module namespace objects. In ES6, a namespace object is never callable, so e.g. `import * as assert from "assert"` will never work as you want. The fix was to enable the TS flags `esModuleInterop`, which makes TS complain about non-standard module semantics, and `allowSyntheticDefaultImports` which shims commonjs require to allow importing commonjs modules as default, so `import assert from "assert"` instead.
* Rely less on node module resolution semantics. Rollup understands some things thanks to the local-resolve plugin, but importing directories didn't work well. I've changed them to explicitly importing the index file.
* Rejigger output directories. Typescript now emits es6 modules to `dist-esm`. `dist-esm` is processed further into `dist` by rollup. `dist-test` is also produced by rollup (but should not be included in the production builds).
* With TS emitting es6, ts-node doesn't work (since es6 modules aren't supported natively). The `test-ts` task now overrides this config. This only works on Windows!

There are a couple things that can improve this further that I haven't tackled yet:

1. Use tslib, as there are lots of duplicate helpers in the package
1. Confirm circular deps aren't an issue (or break them :-P) and test the output package further


bterlson:
Rollup is configured to produce a UMD bundle which should work in the browser as before, but I didn't test this either.

christopheranderson:
Dumb question: Does this bundle any of the dependencies? (i.e. we'd be shipping 3rd party dependencies in our pacakge) Or does it just bundle our source?

bterlson:
Not a dumb question! No dependencies are bundled AFAICT. When you run the new pack command it will tell you a few things about what it's doing. External deps are still loaded with require in Node, and presumably need to exist as a "UMD" module in the browser. I imagine this can be configured if it's not to your liking.

christopheranderson:
Does this produce source maps so we can debug the "production" package properly?

bterlson:
It does now 😄

DanielRosenwasser:
> sorry Sean

It's actually the official advice for libraries 😃

https://medium.com/webpack/webpack-and-rollup-the-same-but-different-a41ad427058c

southpolesteve:
+1 to rollup. 

I don't think we are actually shipping the webpack'd bundle. main points to `lib/index.js`. Webpack there just there to ensure it works (at least superficially) for browser users. Any thoughts about the best way to keep that? We could run it on CI using `npx`

bterlson:
My belief is that the `dist/index.js` bundle in this PR should work in Node and the browser without modification. If this isn't desired, rollup can be configured to produce a commonjs package instead of UMD and yeah we can produce the UMD build as-needed.

bterlson:
Also fwiw there is a rollup plugin to do similar analysis as #103 added to webpack.

amarzavery:
@RikkiGibson @daschult - Please take a look at this PR. It can be helpful for bundling browser modules  that would be very light weight.

RikkiGibson:
The time-to-require refers to the nodejs environment, correct? Is this a case where we can improve the experience for node as well as bundler users by shipping an index.js that's been minified and shaken ahead of time?

bterlson:
@RikkiGibson time-to-require refers to Node, yeah.  Basically benchmarked a single-line script that calls `require('cosmos')` for both flattened and unflattened.

I think this is an improvement for people who aren't bundling (or haven't bundled yet, because development). Consumers doing bundling will use the `dist-esm` (pointed to by `module`) to do tree shaking. Since the bundle is UMD, it can also be  script-included directly from node_modules into the browser for users who find that convenient - in that case, again, the overhead should be much lower than the previous webpack build.

Also keep in mind that the optimization here isn't minification per se, it's flattening. Flattening just crams together a bunch of source files into one file. I tested with minification in addition to flattening (default uglify-es setup) and didn't see any improvements.

Realizing there are more todos here:

* Package.json needs the module entry of `dist-esm/index.js`
* Main is no longer pointing to the right place
* clean task and maybe others need updates.

RikkiGibson:
OK, I think that makes sense.. So the file referenced in `package.main` is gonna be the flattened module, right? And in practice, that's probably the only artifact you want to distribute in commonjs format?

bterlson:
pkg.main/`dist` should point to the output of rollup and yeah should be the only commonjs artifact (although in this case it's UMD so you can also script include it directly into the  browser).

pkg.module/`dist-esm` should take care of the tooling scenarios (bundling and what-have-you).

The Angular folk ship multiple flavors of esm at various levels of transpilation and flatness. In older versions of webpack it was apparently best for your app's package size if your library's pkg.module was flattened such that things that would tree-shake together were flattened together (so if you had two main exports, you'd have two flattened modules). Since this is rapidly becoming unnecessary my feeling is we don't need the added complexity.

southpolesteve:
@bterlson makes sense. UMD 👍. 

I still need to make sure we're not forcing massive webpack bundles on downstream users via polyfills (👀 `crypto`) but that's outside this PR.

bterlson:
I was actually pretty shocked at how little raw size mattered for load times. I see no appreciable difference in require times between the minified version of this library (~123kb) with the unminified version (~390kb). For node users, the impact of shipping a polyfill (even if it won't be needed) might not be that bad in practice as long as its flattened in with the rest of the code.

bterlson:
Package.json should now point to all the right places and the tasks should work. I've also added the `module` entry pointing to `dist-esm`.

A package.json field of `sideEffects: false` will let Webpack optimize this library even further, but can only be applied if none of the modules in this package have side effects as a result of their evaluation (e.g. they shim something that is used later). I'll leave it to @christopheranderson to advise whether this flag should be added 😊

christopheranderson:
For Node.js users, the file size mostly doesn't matter. Even many MBs will load pretty danged fast. It's a bigger issue with download times/etc.

As I understand it, it's the browser scenario that size matters, as well as enabling async loading.

package.json
```diff
@@ -20,18 +20,16 @@
...
     "check-format": "prettier --list-different --config .prettierrc.json \"src/**/*.ts\"",
     "compile": "echo Using TypeScript && tsc --version && tsc --pretty",
     "docs": "typedoc --excludePrivate --exclude \"**/test/**\" --mode file --out ./dist/docs ./src",
     "pack": "rollup -c",
     "build": "npm run clean && npm run check-format && npm run lint && npm run compile && npm run docs && npm run pack",
-    "test": "mocha -r ./src/test/common/setup.ts ./dist-test/index.js --timeout 100000 -i",
-    "test-ts": "set TS_NODE_COMPILER_OPTIONS={\"module\":\"commonjs\"} && mocha -r ts-node/register -r ./src/test/common/setup.ts ./src/test/**/*.spec.ts --recursive --timeout 100000 -i -g .*ignore.js",
-    "test-browser": "karma start ./karma.config.js --single-run",
-    "tslint-check": "tslint-config-prettier-check ./tslint.json"
+    "test": "cross-env TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' mocha -r ts-node/register -r ./src/test/common/setup.ts ./src/test/**/*.spec.ts --recursive --timeout 100000 -i -g .*ignore.js",
```
southpolesteve: I made this cross platform happy with `cross-env`and removed any separate build targets for test. Only one test command now.

src/CosmosClient.ts
```diff
@@ -1,15 +1,14 @@
 import { Agent, AgentOptions } from "https";
 import * as tunnel from "tunnel";
 import * as url from "url";
-import { Constants, RequestOptions } from "./index";
```
southpolesteve: Changed due to circular dep issue

tsconfig.prod.json
```diff
@@ -1,6 +0,0 @@
-{
```
southpolesteve: No longer needed

rauchg:
I'm curious: how come webpack's bundle is 10 times slower to require? Doesn't sound right to me. cc @thelarkinn @wsokra

sokra:
I read the times as bundled vs unbundled.

bterlson:
@rauchg the webpack config prior to this PR was generating one file per module (mirroring the input source) rather than bundling into one .js file. Going from many files to one file has at least 10x improvements in time-to-first-require (on my machine at least).

southpolesteve:
@christopheranderson This is ready for review. Tests are not running here because of when it was opened, but they pass for me locally (including the ts consumer test)

src/common/constants.ts
```diff
@@ -178,7 +176,7 @@ export const Constants = {
   CurrentVersion: "2018-06-18",
 
   SDKName: "azure-cosmos-js",
-  SDKVersion: version,
+  SDKVersion: process.env.VERSION,
```
christopheranderson: I'm guessing this is coming from Rollup? Can we have a PREFIX to make it clear this is coming from rollup and not something we expect users to set?

southpolesteve: Discussed with @christopheranderson offline. We'll need to use the approach on master to make this work in all environments. ESM (webpack) consumers won't get the SDK version set properly with rollup find/replace.

PR #106, opened 2018-08-22
Multi-Region Write support

christopheranderson:
This isn't ready to merge quite yet, but I'm opening the PR because it's just a few bugs left.

Changes:
1. [BREAKING] Session token format changed
2. [BREAKING] Rolls the API version forward
3. Conflict types added
4. Adds MRW to the Connection Policy
5. Changes the Global Endpoint manager to support write region changes
   - Also supports background refresh process and some other failover goodies
   - Adds location cache
6. Refactors Session Container massively to be clearer on how it is working/etc.

### DIFF
--- a/src/LocationCache.ts
+++ b/src/LocationCache.ts
@@ -1,5 +1,5 @@
 import { CosmosClientOptions } from "./CosmosClientOptions";
-import { DatabaseAccount } from "./documents";
+import { DatabaseAccount, Location } from "./documents";
 
 enum EndpointOperationType {
   None = "None",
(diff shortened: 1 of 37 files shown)


src/LocationCache.ts
```diff
@@ -0,0 +1,321 @@
...
+      unavailabilityInfo == null ||
+      !unavailabilityInfo.operationTypes.has(expectedAvailableOperations)
+    ) {
+      return false;
+    } else {
+      if (
+        new Date(Date.now() - this.options.connectionPolicy.backgroundRefreshLocationTimeIntervalMS) >
+        unavailabilityInfo.lastUnavailablityCheckTimeStamp
+      ) {
+        return false;
```
christopheranderson: move this logic into a method

src/retry/endpointDiscoveryRetryPolicy.ts
```diff
@@ -32,19 +33,34 @@ export class EndpointDiscoveryRetryPolicy {
...
+    this.currentRetryAttemptCount++;
+
+    if (Helper.isReadRequest(this.request)) {
+      this.globalEndpointManager.markCurrentLocationUnavailableForRead();
+    } else {
+      this.globalEndpointManager.markCurrentLocationUnavailableForWrite();
+    }
+
+    if (!Helper.isReadRequest(this.request) && this.globalEndpointManager.getAlternateEndpoint()) {
+      this.request.retryCount++;
```
christopheranderson: should do this in request instead (or else this will only increment in the endpointDiscoveryPolicy)

christopheranderson:
I'm gonna update travis to use the new Multi-master endpoint and VSTS to use the new Container version, which has the new API version supported. This won't work on any "old" emulators/etc.

PR #107, opened 2018-08-22
Replace crypto with create-hmac package

southpolesteve:
This decreases the webpacked size by ~1mb

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -1,4 +1,4 @@
-import * as crypto from "crypto";
+import createHmac = require("create-hmac");
 import { PermissionDefinition } from "./client";
 import { Helper } from "./common";
 import { IHeaders } from "./queryExecutionContext";
@@ -95,8 +95,7 @@ export class AuthHandler {
       "\n";
 
     const body = new Buffer(text, "utf8");
-    const signature = crypto
-      .createHmac("sha256", key)
+    const signature = createHmac("sha256", key)
       .update(body)
       .digest("base64");
     const MasterToken = "master";
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -1,4 +1,4 @@
-import createHmac = require("create-hmac");
+import createHmac from "create-hmac";
 import { PermissionDefinition } from "./client";
 import { Helper } from "./common";
 import { IHeaders } from "./queryExecutionContext";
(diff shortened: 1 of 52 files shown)


src/auth.ts
```diff
@@ -1,4 +1,4 @@
-import * as crypto from "crypto";
+import createHmac = require("create-hmac");
```
southpolesteve: @christopheranderson is this best way to do this? This module is distributed as a single exported function.

src/typings/create-hmac.d.ts
```diff
@@ -0,0 +1,15 @@
+declare module "create-hmac" {
```
southpolesteve: Pulled these from the node typings. The npm module has the same API.

src/auth.ts
christopheranderson: `import * as createHmac from "create-hmac"`

christopheranderson: If they do "default" exports (which they probably don't), you could do `import createHmac from "create-hmac"`, but I'd guess they don't do that.

christopheranderson: Could we abstract away the hmac logic into a separate file, which does the browser detection stuff? I'd prefer to use crypto with Node.js to remain standard there. I'm fine with create-hmac for Browser.

southpolesteve: When I use `import * as createHmac` I get `[ts] Module '"create-hmac"' resolves to a non-module entity and cannot be imported using this construct.` from typescript

southpolesteve:
That is exactly what the package does if running in node https://github.com/crypto-browserify/createHmac/blob/master/index.js

src/auth.ts
christopheranderson: Ah, this was a nasty one.

So you can add this to the tsconfig.json
```json
        // Enable support for importing CommonJS modules targeting es6 modules
        "esModuleInterop": true,

        // When using above interop will get missing default export error from type check since
        // modules use "export =" instead of "export default", enable this to ignore errors.
        "allowSyntheticDefaultImports": true
```

Then you can import it like a default import

```typescript
import createHmac from "create-hmac";
const hmac = createHmac("sha256", "hello world").update("I love cupcakes").digest("hex");
console.log(hmac);
```

Gist as a proof of concept: https://gist.github.com/christopheranderson/e3f0f0777e8a80919776a3ef5d8a4d6f

Issue #61, opened 2018-07-20
Browser Compatibility


southpolesteve:
This is a meta issue to track other work and questions around browser compatibility. One of our goals is to make the SDK useable in the browser and we still have some work to do. The lib currently webpack'd as part of the build step but not much has been done to verify it is usable in browsers.

Current open questions:
- Does it *actually* work in all major browsers?
- Can we get tests running in the browser and part of CI?
- What is the current best practice around shipping npm packages for both node and browsers?
- Can our lib be easily consumed by downstream webpack or rollup bundles?
- Can we make our bundled size reasonable? Current `webpack  -p`: 3.17 MiB 😱

southpolesteve:
v2 is compatible. v3 is extra awesomely compatible because it is based on fetch internals

PR #114, opened 2018-09-03
Remove sourcemap from the bundle in webpack -p mode. Helps #61

alippai:
This way it's only 323kB (86kB gzipped)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9601,8 +9601,7 @@
     "tslib": {
       "version": "1.9.3",
       "resolved": "https://registry.npmjs.org/tslib/-/tslib-1.9.3.tgz",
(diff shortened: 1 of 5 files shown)


alippai:
After running `npm run build-prod` instead of `npm run build` it's 279kB (74kB gzipped)

alippai:
@christopheranderson This is more like a bugfix for the current RC (as a temporary solution) while #104 (nice!) lands.

I didn't notice the PR (#104) before, just the issue (#61) and I thought this is something I can improve. Feel free to close this PR :)

| What | Time |
|--------------------------|--------|
| Loading the lib/index.js | >150ms |
| Loading the old lib/dist/azurecosmos.js | >100ms |
| Loading the new lib/dist/azurecosmos.js produced with the new build-prod | ~40ms |

southpolesteve:
@alippai Thank you for this PR! Wanted to let you know I'll be looking at it soon. I need to have a larger discussion with the team and how it fits in with #104. Hopefully this next week.

southpolesteve:
@alippai Thanks for this! Going to merge it in now. We may make some tweaks once #104 lands, but at least right now this is helpful as we try to ensure the SDK is slim for browser users.

PR #126, opened 2018-09-18
Multi-Region Write support (#106)

christopheranderson:


christopheranderson:
All the contents of this PR are already approved, and all our tests pass now that the emulator is updated to support latest API version, so FULL STEAM AHEAD! 🚢 🚢 🚢

PR #128, opened 2018-09-19
Remove unused npm script

southpolesteve:


christopheranderson:
I think the stage environment makes the ttl tests take a long time. Your code doesn't touch them. :shipit:

PR #129, opened 2018-09-19
Pull in all of esnext lib

### LINKED ISSUE #127: Cannot find name 'AsyncIterable'
The following error is thrown

```
c:/dev/burkeholland/ignite-app/node_modules/@azure/cosmos/lib/queryIterator.d.ts(68,25): error TS2304: Cannot find name 'AsyncIterable'.
```

This is resolved by adding the following to `compilerOptions` in the tsconfig.json...

```
"lib": [
  "esnext"
]
```

Perhaps we can mention something in the docs about this? Or maybe I'm doing something else wrong?

southpolesteve:
Fixes #127. I'm not exactly sure why this is necessary, but seeing reports of a similar issue in another lib https://github.com/apollographql/graphql-subscriptions/issues/83

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,7 +13,7 @@
     "target": "es6",
     "sourceMap": true,
     "newLine": "LF",
-    "lib": ["es6", "esnext.asynciterable"]
+    "lib": ["es6", "esnext"]
   },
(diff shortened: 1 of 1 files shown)


christopheranderson:
I think the stage environment makes the ttl tests take a long time. Your code doesn't touch them. :shipit:

PR #130, opened 2018-09-20
Fix/mrw fixes

christopheranderson:
Last minute fixes for multi-region write


samples/MultiRegionWrite/ConflictWorker.ts
```diff
@@ -125,6 +125,7 @@ export class ConflictWorker {
   }
 
   private async RunUpdateConflictOnManual() {
+    let retryCount = 5;
```
southpolesteve: Why is this 5?

christopheranderson: Arbitrary retry count for the sample

PR #146, opened 2018-09-25
Update README.md

southpolesteve:
Fix links that previously pointed to preview

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,8 @@ helloCosmos().catch(err => {
 ## Useful links
 
 - [Welcome to Azure Cosmos DB](https://docs.microsoft.com/en-us/azure/cosmos-db/community)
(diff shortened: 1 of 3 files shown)


christopheranderson:
We should do this into dev first, or does this need to be hotfixed into master?

southpolesteve:
Master. The links 404 on master right now

christopheranderson:
It'll 404 on npm as well. We need to do a full patch release to update npm, right?

southpolesteve:
Yes

christopheranderson:
Maybe just update this PR with the package update info, then? We'll need to cherry pick this back onto dev, as well.

southpolesteve:
@christopheranderson Bumped the version in here. As soon as this passes it can be merged and published to npm.

christopheranderson:
Adding that version checking test looks like it was worth it :)

southpolesteve:
This will not be the last time that test bites me

PR #161, opened 2018-10-20
Remove lib from tsconfig. Use reference directive for AsyncIterable

southpolesteve:
Shipping a library that specifies any built-in TS lib types is problematic as downstream consumers will also have to add that lib option to their tsconfig. We can work around this by using a triple-Slash directive instead. See https://github.com/Microsoft/TypeScript/issues/27416

@burkeholland I believe this should fix https://github.com/Azure/azure-cosmos-js/issues/127 without you having to specify `lib: ["esnext"]` in your own repo

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@azure/cosmos",
-  "version": "2.0.0",
+  "version": "2.0.2",
   "lockfileVersion": 1,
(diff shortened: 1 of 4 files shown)


southpolesteve:
@bterlson We should chat about this. I think it will be important for other libraries not to run into this same issue

PR #162, opened 2018-10-21
typo: teh -> the

nschonni:
Closes https://github.com/MicrosoftDocs/azure-docs-sdk-node/pull/372

PR #164, opened 2018-10-21
typo: Double word "the"

nschonni:
Also changed to "will before" -> "will be before"


southpolesteve:
Pulled in as part of https://github.com/Azure/azure-cosmos-js/pull/162

PR #174, opened 2018-10-31
Enable no-uncessary-class tslint rule

southpolesteve:
I need to double check we are still exporting everything we were before. Otherwise, this might be a breaking change.


christopheranderson:
We should point this to v3 and merge it or close it.

southpolesteve:
I'll redo this on top of v3 once #104 Lands

southpolesteve:
Closing in favor of #211

Issue #160, opened 2018-10-20
Add production dependency on @types/node


southpolesteve:
Typescript users of this library need to install @types/node. Guidance from Typescript docs say we should include it in dependencies: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

Upside: 
- Better out of the box experience for TS users

Downside: 
- Our version of node types could conflict with a user's version?
- Non-TS users download unnecessary packages

christopheranderson:
Another alternative to weigh here to to get rid of these types all together. I think we're doing this just for the `Agent` type. I wonder if we could just create an interface for that type and remove the dependency all together.

christopheranderson:
Overall, though, I don't have a problem with the downsides except some worry about the version conflicts, which are always annoying. We'll want to target the lowest LTS we support.

southpolesteve:
Closed by https://github.com/Azure/azure-cosmos-js/pull/175. We no longer expose types from @types/node

PR #175, opened 2018-11-01
Mirror node Agent interface. Prevents leaking of node types

southpolesteve:
Confirmed using https://github.com/Microsoft/web-build-tools/wiki/API-Extractor that this change no longer emits a reference to node types 🎉

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@azure/cosmos",
-  "version": "2.0.3",
+  "version": "2.0.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


southpolesteve:
@christopheranderson added

southpolesteve:
Closes #160

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@azure/cosmos",
-  "version": "2.0.3",
+  "version": "2.0.4",
```
christopheranderson: Did we just forget to check this in last time?

christopheranderson: We should add to the version checking test to also look in package-lock.json to make sure we don't forget it.

southpolesteve: 🤦‍♂️ Seems like npm should not allow this....

I may just write our own publish script. I keep running into things like this.

PR #176, opened 2018-11-01
support for ChangeFeedPolicy

srinathnarayanan:
This PR
- Adds MultiPolygon Datatype to JS SDK
- Adds support for OperationLogPolicy to be set in collection definition

### DIFF
--- a/src/client/Container/ContainerDefinition.ts
+++ b/src/client/Container/ContainerDefinition.ts
@@ -1,4 +1,4 @@
-import { IndexingPolicy, OperationLogPolicy, PartitionKey, PartitionKeyDefinition } from "../../documents";
+import { ChangeFeedPolicy, IndexingPolicy, PartitionKey, PartitionKeyDefinition } from "../../documents";
 import { ConflictResolutionPolicy } from "../Conflict/ConflictResolutionPolicy";
 
 export interface ContainerDefinition {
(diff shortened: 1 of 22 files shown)


src/client/Container/ContainerDefinition.ts
```diff
@@ -12,4 +12,6 @@ export interface ContainerDefinition {
   defaultTtl?: number;
   /** The conflict resolution policy used to resolve conflicts in a container. */
   conflictResolutionPolicy?: ConflictResolutionPolicy;
+  /** The conflict resolution policy used to resolve conflicts in a container. */
```
southpolesteve: Is this comment correct?

src/client/Container/ContainerDefinition.ts
```diff
@@ -12,4 +12,6 @@ export interface ContainerDefinition {
   defaultTtl?: number;
   /** The conflict resolution policy used to resolve conflicts in a container. */
   conflictResolutionPolicy?: ConflictResolutionPolicy;
+  /** The opearation log policy for a container. */
```
christopheranderson: nit: this isn't a descriptive comment, imo. Something like "The operation log policy for a container determines the log retention period for your change feed" (not sure if that's correct)

christopheranderson:
Don't merge until we've gotten an Ack we're actually shipping it

PR #189, opened 2018-11-13
Allow override of createRequestObject

southpolesteve:


southpolesteve:
Not going to go this route

PR #194, opened 2018-11-16
Install and Build during Prepack step

### LINKED ISSUE #192: Installable from Github
You can't currently install from github: `npm install Azure/azure-cosmos-js`

We need to hook into the correct npm scripts (prepack?) to build on git install.

southpolesteve:
Closes #192. This will only install devDependencies if you install from Github. Normal installs from npm are unaffected. Has the added benefit of making sure we never ship code built with out of date dependencies to Github.


southpolesteve:
Welp apparently this also infinite loops CI

christopheranderson:
Does prepack get called by npm when it is installing via GitHub?

southpolesteve:
Yep!

PR #196, opened 2018-11-26
Add change feed support

christopheranderson:
Changes:
- Adds readChangeFeed method to Items
- Adds ChangeFeedIterator implementation + associated interfaces
- Makes readPartitionKeyRanges public on Container + associated interfaces
- Adds status code to Response
- completely changed the incrementalfeed test to use the new change feed iterator type
- Adds retry to Session tests (flaky)

Todo:
- [x] Add samples
- [ ] ChangeFeedResponse unit tests


src/ChangeFeedIterator.ts
```diff
@@ -0,0 +1,116 @@
+/// <reference lib="esnext.asynciterable" />
+import { isNumber } from "util";
```
southpolesteve: I wouldn't block this PR for it but removing all uses of `util` is on my radar for decreasing the bundle size

src/ChangeFeedIterator.ts
```diff
@@ -0,0 +1,116 @@
...
+  constructor(
+    private clientContext: ClientContext,
+    private resourceId: string,
+    private resourceLink: string,
+    private isPartitionedContainer: () => Promise<boolean>,
+    private changeFeedOptions: ChangeFeedOptions
+  ) {
+    // partition key XOR partition key range id
+    const partitionKeyValid = changeFeedOptions.partitionKey !== undefined;
+    const partitionKeyRangeIdValid = !(
```
southpolesteve: What's going on with this check? Does it mean passing partitionKeyRangeId === false is valid?

christopheranderson: That would be invalid, since it would be an invalid type. I should probably add a type check that it is a string. That would actually cover the other two cases, as well.

src/ChangeFeedIterator.ts
christopheranderson: ACK, but I'm gonna leave this as-is since I don't want to implement and write all those tests this week. When it comes time for you to remove this, you can ask me to do that, if you'd like.

christopheranderson:
@southpolesteve - could you re-review the latest commit? I changed the response type and I'd like some feedback.

src/ChangeFeedIterator.ts
```diff
@@ -101,14 +129,21 @@ export class ChangeFeedIterator<T> {
...
       this.resourceId,
       result => (result ? result.Documents : []),
       undefined,
       feedOptions,
       this.changeFeedOptions.partitionKeyRangeId
-    ) as Promise<any>; // TODO: some funky issues with query feed. Probably need to change it up.
+    ) as Promise<any>); // TODO: some funky issues with query feed. Probably need to change it up.
+
+    return new ChangeFeedResponse(
+      response.result,
```
southpolesteve: any reason these params need to be split up? Could just pass `response` and keep all this logic inside  `ChangeFeedResponse

southpolesteve:
Unclear on one thing. Would this introduce a mismatch between our different iterators? One would return `result` and the change feed iterator `results`?

src/ChangeFeedIterator.ts
christopheranderson: Call it personal preference. I like to make it clear what an object requires. I like being able to pass a null result or something, in the future, without having to mutate the response object.

Also, I like using the `public` syntax on objects, as well, so I don't need to unpack in the constructor.

Thirdly, I don't like the current `Response` object, and I want to get rid of it in v3, so it's one less object I've gotta touch.

southpolesteve: Got it. You've convinced me.

christopheranderson:
Yeah, I was wondering if you'd find that weird. I like that results implies it will be an array (via the plural), where as result is not so clear. But is that distinction worth the confusion? I'll let you be the final arbiter on that. 

I also considered naming it "changes" since that's what this response type is.

southpolesteve:
I like `changes`. I also like switching to `results` for all potential arrays. But that would be breaking. If we went with `changes` now it would be another weird change from v2 to v3 to switch to `results`.... Ugh. Let's go with `results` for now if we plan to switch to `results` for all array returns in v3.

christopheranderson:
Would it make you feel better if I had "result" and "results"? I can add another getter which does that.

southpolesteve:
I'm cool with "result" and "results". Don't feel like implementing Response is necessary if you wanna change it in v3

src/client/Item/Items.ts
```diff
@@ -72,6 +81,72 @@ export class Items {
     return new QueryIterator(this.clientContext, query, options, fetchFunction, this.container.url);
   }
 
+  /**
+   * Create a `ChangeFeedIterator` to iterate over pages of changes
+   * @param partitionKey
+   * @param changeFeedOptions
+   */
+  public readChangeFeed(
+    partitionKey: string | number | boolean,
```
mkolt: Why is partitionKey a parameter here? It's part of changeFeedOptions, right?

src/client/Item/Items.ts
```diff
@@ -72,6 +81,72 @@ export class Items {
...
+    changeFeedOptions: ChangeFeedOptions
+  ): ChangeFeedIterator<T>;
+  /**
+   *
+   * @param changeFeedOptions
+   */
+  public readChangeFeed<T>(changeFeedOptions?: ChangeFeedOptions): ChangeFeedIterator<T>;
+  public readChangeFeed<T>(
+    partitionKeyOrChangeFeedOptions?: string | number | boolean | ChangeFeedOptions,
```
mkolt: I don't think we need that. Why don't we just have changeFeedOptions?

src/client/Item/Items.ts
```diff
@@ -72,6 +81,72 @@ export class Items {
...
+    }
+
+    const path = Helper.getPathFromLink(this.container.url, "docs");
+    const id = Helper.getIdFromLink(this.container.url);
+    return new ChangeFeedIterator<T>(
+      this.clientContext,
+      id,
+      path,
+      async () => {
+        return !!(await this.container.getPartitionKeyDefinition()).body;
```
mkolt: nit: !! looks a bit cryptic, can we use something like != null instead?

src/client/Item/Items.ts
christopheranderson: Intentional choice to make users think about partition key. We've been doing this for the new .NET SDK we have we overloads for all the Item APIs as well. @southpolesteve, do you agree?

src/ChangeFeedIterator.ts
```diff
@@ -0,0 +1,149 @@
...
+      feedOptions.maxItemCount = this.changeFeedOptions.maxItemCount;
+    }
+
+    if (this.changeFeedOptions.sessionToken) {
+      feedOptions.sessionToken = this.changeFeedOptions.sessionToken;
+    }
+
+    if (this.nextIfNoneMatch) {
+      feedOptions.accessCondition = {
+        type: "IfNoneMatch",
```
mkolt: use HttpHeaders.IfNoneMatch.

src/client/Item/Items.ts
```diff
@@ -72,6 +81,72 @@ export class Items {
...
+      changeFeedOptions = partitionKeyOrChangeFeedOptions;
+    } else if (partitionKeyOrChangeFeedOptions !== undefined && !isChangeFeedOptions(partitionKeyOrChangeFeedOptions)) {
+      partitionKey = partitionKeyOrChangeFeedOptions;
+    }
+
+    if (!changeFeedOptions) {
+      throw new Error("changeFeedOptions must be a valid object");
+    }
+
+    if (partitionKey !== undefined) {
```
mkolt: not sure why we need all this complication of partition key in 2 places

src/test/integration/incrementalFeed.spec.ts
```diff
@@ -1,106 +1,501 @@
...
+      // create container and two items
+      before(async function() {
+        container = await getTestContainer("Newly updated items should be fetched incrementally");
+      });
+
+      after(async function() {
+        await container.delete();
+      });
+
+      // TODO: start time is currently being ignored. Need to investigate.
```
mkolt: Is this closed now or still something is not working? If cloesed, let's remove the comment.

src/client/Item/Items.ts
```diff
@@ -72,6 +81,72 @@ export class Items {
     return new QueryIterator(this.clientContext, query, options, fetchFunction, this.container.url);
   }
 
+  /**
+   * Create a `ChangeFeedIterator` to iterate over pages of changes
+   * @param partitionKey
+   * @param changeFeedOptions
+   */
+  public readChangeFeed(
```
mkolt: usage example would be good to have. We generate jsdoc from this, right?

src/ChangeFeedIterator.ts
```diff
@@ -0,0 +1,149 @@
...
+import { Resource } from "./client";
+import { ClientContext } from "./ClientContext";
+import { Constants, ResourceType, StatusCodes } from "./common";
+import { FeedOptions } from "./request";
+import { Response } from "./request";
+
+/**
+ * Provides iterator for change feed.
+ *
+ * Use `Items.readChangeFeed` to get an instance of the iterator.
```
mkolt: Shuld this be Items.readChangeFeed() ?

src/client/Item/Items.ts
christopheranderson: I can add a code sample here, yeah. It will show up in our reference docs.

src/test/integration/incrementalFeed.spec.ts
christopheranderson: it's working now, yeah. Good catch!

src/ChangeFeedIterator.ts
```diff
@@ -0,0 +1,149 @@
...
+ */
+export class ChangeFeedIterator<T> {
```
mkolt: Why do we call this Iterator? It doesn't have the "next" method. As it is, it's more like a ChangeFeedQuery in C# SDK. Like Iterable but without [Symbol.iterator]/[Symbol.asyncIterator] member, and still "query" makes more sense for naming as it has more stuf than just the iterator. Note that I can't use it with for..of. Why not? here's what would be very natural to do in JS:
for await (var item Items.readChangeFeed(options)) process(item);

src/client/Item/Items.ts
mkolt: Yes, please, would make big difference.

src/client/Item/Items.ts
mkolt: well, looks quite messy. Do we really have to go this way? :)

src/ChangeFeedIterator.ts
christopheranderson: The way we've implemented the async iterator, it is backward compatible (but folks will only be able to use `for await` in the newer versions of Node.js or with babel/typescript polyfills). We already use this async Iterator pattern in the `QueryIterator` type.

Query returns a `QueryIterator`, so it made sense to me that change feed returns a `ChangeFeedIterator`. `ChangeFeedQuery` isn't consistent with our existing naming in the JS SDK.

FWIW, I'd prefer `*Cursor` for both to avoid the `Iterator` confusion, but not confuse SQL queries, read feed with change feed.

src/client/Item/Items.ts
southpolesteve: PartitionKey will soon be required right? If there are one or two required things in a function call I like supporting the overload. Makes the beginner experience clearer.

christopheranderson: re: being messy - isn't gonna be messy since JavaScript doesn't officially support overloads, so we have to handle all those overloads with 1 implementation. Just a reality of JavaScript, unfortunately. Other option is to always make them required and pass undefined in, which is a worse customer experience.

src/client/Item/Items.ts
christopheranderson: !! is valid JavaScript for seeing if an object is truthy (aka not undefined or null). Agree this is confusing to read at first, though, so I refactored into two lines with some meaningful variable names

christopheranderson:
Feedback results:
- Remove Partition Key from Options
- Hide PartitionKeyRange on Options
- Iterator name is fine

PR #198, opened 2018-11-26
Remove `Stream` as possible body type

southpolesteve:
@christopheranderson Let's chat about this today. I don't believe we allow passing a stream anymore anywhere so this should be a non-breaking change.


christopheranderson:
This should be safe. We definitely don't expose streams anywhere.

PR #200, opened 2018-12-02
Use npx to install typescript for ts-test

southpolesteve:


southpolesteve:
npx should exist if you have npm > 5.2 installed which should be the case for anyone testing this lib or on CI

PR #201, opened 2018-12-03
Remove assert and util in lib code

southpolesteve:
Added a tslint rule so we don't accidentally import them again


christopheranderson:
Why are we getting rid of assert?

southpolesteve:
Decreasing bundled size. My main goal here was actually `util`. But `assert` came along for the ride. If we really want it, the polyfill is tiny: https://bundlephobia.com/result?p=assert@1.4.1

PR #206, opened 2018-12-09
Fix typoes

orta:
Spotted some while reading the docs, figured I should just look out for a few overall.

### DIFF
--- a/src/client/Item/ItemDefinition.ts
+++ b/src/client/Item/ItemDefinition.ts
@@ -1,7 +1,7 @@
 /**
  * Items in Cosmos DB are simply JSON objects.
  * Most of the Item operations allow for your to provide your own type
- * that extends the very simple Itemdefinition.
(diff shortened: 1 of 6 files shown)


src/client/Item/ItemDefinition.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * Items in Cosmos DB are simply JSON objects.
  * Most of the Item operations allow for your to provide your own type
- * that extends the very simple ItemDefintion.
+ * that extends the very simple Itemdefinition.
```
southpolesteve:
```suggestion
 * that extends the very simple ItemDefinition.
```
I think we want to keep this one the same to match the type name

southpolesteve:
Thank you! 🙌

src/client/Item/ItemDefinition.ts
orta: Good thinking!

PR #207, opened 2018-12-11
Fix EnableEndpointDiscovery

southpolesteve:
In the current code we still make the request, but ignore the results. This makes it so the request never happens.


christopheranderson:
Is there an obvious place to add test coverage for this flag?

PR #210, opened 2018-12-17
Remove tunnel

southpolesteve:
In v3 we'll support proxies via a custom HTTP(s) agent such as https://github.com/TooTallNate/node-proxy-agent. This means we can remove tunnel and ProxyUrl as a constructor option.


christopheranderson:
Looks good. Only worry is that these tests don't run in CI, so can't be 100% sure this is all good. Will also need to follow up with some documentation updates once v3 is GA'd.

Created this item to track turning the tests on in CI: https://github.com/Azure/azure-cosmos-js/issues/212

Issue #32, opened 2018-07-06
Enable no-unnecessary-class lint rule


southpolesteve:
Seems to me like a good thing https://palantir.github.io/tslint/rules/no-unnecessary-class/

christopheranderson:
Do you have an example of where we have unnecessary classes today?

Not opposed to this, just wondering where it came up?

southpolesteve:
Came up while I was making prettier and tslint play nice. Here is where it throws errors right now:

```
ERROR: /Users/steve/azure-cosmos-js/src/auth.ts[18, 8]: Every member of this class is static. Use namespaces or plain objects instead.
ERROR: /Users/steve/azure-cosmos-js/src/base.ts[10, 8]: Every member of this class is static. Use namespaces or plain objects instead.
ERROR: /Users/steve/azure-cosmos-js/src/common/helper.ts[5, 8]: Every member of this class is static. Use namespaces or plain objects instead.
ERROR: /Users/steve/azure-cosmos-js/src/common/platform.ts[4, 8]: Every member of this class is static. Use namespaces or plain objects instead.
ERROR: /Users/steve/azure-cosmos-js/src/common/uriFactory.ts[3, 8]: Every member of this class is static. Use namespaces or plain objects instead.
ERROR: /Users/steve/azure-cosmos-js/src/hash/murmurHash.ts[4, 8]: Every member of this class is static. Use namespaces or plain objects instead.
ERROR: /Users/steve/azure-cosmos-js/src/queryExecutionContext/headerUtils.ts[11, 8]: Every member of this class is static. Use namespaces or plain objects instead.
ERROR: /Users/steve/azure-cosmos-js/src/queryExecutionContext/partitionedQueryExecutionContextInfoParser.ts[18, 8]: Every member of this class is static. Use namespaces or plain objects instead.
ERROR: /Users/steve/azure-cosmos-js/src/queryMetrics/queryMetricsUtils.ts[4, 8]: Every member of this class is static. Use namespaces or plain objects instead.
ERROR: /Users/steve/azure-cosmos-js/src/request/request.ts[143, 8]: Every member of this class is static. Use namespaces or plain objects instead.
ERROR: /Users/steve/azure-cosmos-js/src/retry/retryUtility.ts[22, 8]: Every member of this class is static. Use namespaces or plain objects instead.
[...]

PR #211, opened 2018-12-17
Enable no-unecessary-class (v3)

southpolesteve:
Supersedes #32


src/tslint.json
```diff
@@ -0,0 +1,5 @@
+{
```
southpolesteve: Idk how this got in here. Removing

PR #213, opened 2018-12-19
Generate tokens using @azure/cosmos-sign (v3)

southpolesteve:
I already did a bunch of work to minimize the bundle size of token signing in @azure/cosmos-sign so let's take a dep on that package instead of having the code in two places

This also has the benefit of making HTTP methods and Resource types proper enums rather than strings

### DIFF
--- a/src/request/request.ts
+++ b/src/request/request.ts
@@ -4,7 +4,7 @@ import { Socket } from "net";
 import { Stream } from "stream";
 import * as url from "url";
 
-import { Constants, Helper } from "../common";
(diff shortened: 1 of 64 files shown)


src/request/request.ts
```diff
@@ -268,7 +268,7 @@ export async function getHeaders(
     headers[Constants.HttpHeaders.XDate] = new Date().toUTCString();
   }
 
-  if (verb === "post" || verb === "put") {
+  if (verb === "POST" || verb === "PUT") {
```
christopheranderson: Should use constants here since we have them

PR #217, opened 2019-01-08
Refactor tokenProvider and add types

southpolesteve:


src/auth.ts
```diff
@@ -4,18 +4,29 @@ import { Constants, getResourceIdFromPath, HTTPMethod, ResourceType } from "./co
 import { IHeaders } from "./queryExecutionContext";
 
 /** @hidden */
-export interface IRequestInfo {
-  [index: string]: any;
+export interface RequestInfo {
```
christopheranderson: If you're not using Request info in the signature of the method any more you don't need this class, but see my message below about whether we want to do that.

src/auth.ts
```diff
@@ -4,18 +4,29 @@ import { Constants, getResourceIdFromPath, HTTPMethod, ResourceType } from "./co
...
   resourceId: string;
-  resourceType: string;
+  resourceType: ResourceType;
   headers: IHeaders;
 }
 
-export interface ITokenProvider {
-  getToken: (requestInfo: IRequestInfo, callback?: (err: Error, token: string) => void) => Promise<string>;
-}
+export type TokenProvider = (
```
christopheranderson: I'm not a fan of having 6 arguments on a callback in JavaScript. It's harder to consume, imo. That's why I had the wrapper object. What's the value in removing the wrapper?

src/auth.ts
```diff
@@ -4,18 +4,29 @@ import { Constants, getResourceIdFromPath, HTTPMethod, ResourceType } from "./co
...
+  verb: HTTPMethod,
+  path: string,
+  resourceId: string,
+  resourceType: ResourceType,
+  headers: IHeaders,
+  /** The default function for generating header tokens using the masterKey */
+  getAuthorizationTokenUsingMasterKey: (
```
christopheranderson: why are we passing this helper method in each time, not not any of the other helper methods? Shouldn't we just export all these helper methods from a library, and not pass them all in as arguments? This smells weird to me.

src/auth.ts
```diff
@@ -54,25 +68,19 @@ export async function setAuthorizationHeader(
...
-        resourceType,
-        headers
-      })
+      await authOptions.tokenProvider(verb, path, resourceId, resourceType, headers, getTokenUsingMasterKey)
     );
   }
 }
 
-function getAuthorizationTokenUsingMasterKey(
+function getTokenUsingMasterKey(
```
christopheranderson: nit: I don't like this rename. It's less descriptive. We've got like 3-4 different "tokens" in the SDK, and while this is in the "Auth" file, I still don't like not specifying which token.

src/auth.ts
southpolesteve: I think I did this so the call sig matched the others in the file. I can change back though.

src/auth.ts
southpolesteve: Agreed. I'll revisit. This was how the previous token provider worked.

src/auth.ts
christopheranderson: The other ones are internal so I didn't touch them from v1, but I added this one which is why it has the "nicer" model. My preference would be to leave it as a small set of arguments with that wrapper object.

src/auth.ts
southpolesteve: Added this interface back

src/auth.ts
southpolesteve: Changed it to `setAuthorizationTokenHeaderUsingMasterKey` to reflect that it is modifying the header object and exported it from src/index

PR #218, opened 2019-01-15
Extract extractPartitionKey

southpolesteve:
Previously users would have to build up a container reference to use this function. Now they can import it directly from the top level.

Before: 
`client.database(id).container(id).extractPartitionKey(doc, partitionKeyDef)`

after: `extractPartitionKey(doc, partitionKeyDef)`


src/extractPartitionKey.ts
```diff
@@ -0,0 +1,23 @@
+import { parsePath } from "./common";
+import { PartitionKey, PartitionKeyDefinition } from "./documents";
+
+export function extractPartitionKey(document: any, partitionKeyDefinition: PartitionKeyDefinition): PartitionKey[] {
+  // TODO: any
```
christopheranderson: nit: TODO can go away here now

christopheranderson: I probably left this here myself, but as long as we're touching files/etc.

PR #222, opened 2019-01-22
Fix executeNext type

southpolesteve:
- executeNext should return an array inside `result`

### DIFF
--- a/src/queryIterator.ts
+++ b/src/queryIterator.ts
@@ -1,5 +1,4 @@
 /// <reference lib="esnext.asynciterable" />
-import { ItemDefinition, Resource } from "./client";
 import { ClientContext } from "./ClientContext";
 import {
   FetchFunctionCallback,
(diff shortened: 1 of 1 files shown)


src/queryIterator.ts
```diff
@@ -15,7 +16,7 @@ import { Response } from "./request/request";
  * traversal and iterating over the response
  * in the Azure Cosmos DB database service.
  */
-export class QueryIterator<T> {
+export class QueryIterator<T = Resource & ItemDefinition> {
```
southpolesteve: @christopheranderson how about this instead:

`QueryIterator<T = Partial<Resource> & ItemDefinition>`

christopheranderson: Even Partial isn't correct because I can select a concat of properties into a new property. :(

And in this case, Query Iterator can be of type ContainerDefintion/etc. as well, so this really won't work at all. This has to stay T, and then fixed in Items.query where it passes the type along.

southpolesteve:
Also cherry-picked into v3

PR #230, opened 2019-01-27
Demo and demo debugging fixes

AshleyGrant:
1. Set the cwd when using the `Debug File` option in VS Code. This allows simple debugging of the samples. If this breaks debugging scenarios elsewhere, let me know and I can simply create a separate launch config called "Debug Sample File" or something.

2. Documentation tweak to be explicit that the SDK needs to be built before running the samples. It might seem obvious to devs on the team, but I figure being explicit is better than assuming external devs will realize this.

3. For whatever reason, Node refuses to load the `"main"` info from the `package.json` file located in the `lib` folder. This causes the samples to fail to run. I'm fixing this by explicitly pointing to the `lib/src` folder. This is the same place that the `package.json` file in the SDK root points, but I figure for samples, why not be explicit.

4. Typescript typechecking in the JS file finds an error using object literal. Force typescript to see an "any" variable to get rid of this non-error. 

5. Moving to template literals wherever possible.

6. Stylistic move away from for loops in ItemManagement demo. Switch to using `forEach` function on the arrays.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -34,6 +34,7 @@
       "request": "launch",
       "name": "Debug file",
       "program": "${file}",
+      "cwd": "${fileDirname}",
       "env": {
         "NODE_TLS_REJECT_UNAUTHORIZED": "0"
       }
--- a/samples/ChangeFeed/app.js
+++ b/samples/ChangeFeed/app.js
@@ -141,8 +141,8 @@ async function init() {
 }
 
 async function handleError(error) {
-  console.log("\nAn error with code '" + error.code + "' has occurred:");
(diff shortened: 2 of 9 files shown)


AshleyGrant:
I'm not sure why the unit tests failed. I haven't touched any of the SDK code, and all tests run successfully for me locally.

southpolesteve:
@AshleyGrant Don't worry about the tests. We have a bug in the testing infra that is currently causing them to fail. I'll run tests locally when I review this PR. Thanks for sending this!

southpolesteve:
Thanks! Merging now. Will get published in the next release

PR #234, opened 2019-01-29
Add uniqueKeyPolicy to container

southpolesteve:


christopheranderson:
For v3, we'll eventually introduce a builder pattern for this, but this allows portal to be unblocked and easy backporting to master, so this is perfect.

src/client/Container/UniqueKeyPolicy.ts
```diff
@@ -0,0 +1,7 @@
+export interface UniqueKeyPolicy {
```
christopheranderson: We should have TSDocs for public types.

christopheranderson: Don't you need to update ./index.ts to export this type to avoid weird import paths?

southpolesteve: Yep. Added.

PR #238, opened 2019-02-06
Breaking API changes for v3

christopheranderson:
Opening PR for feedback.

This PR has the first wave of the breaking API changes in it.
- Changes all response types to be classes derived from ResourceResponse
- Adds getters for important header values
- Removes nextItem/current from queryIterator
- Changes response type for queryIterator to always be a page of data
- Fixes a bunch of circular dependencies (does not add any testing for them, though)
- Renames IHeaders to CosmosHeaders
- Deletes ruPerMin tests
- Deletes some helper tests
- Changes status codes to object with "as #" names
- Engineering: Adds a common js compile target and vs-code uses that for debugging
- Updates ts-node dependency

Todo:
- [ ] Change all the Options on arguments to be resource/operation specific
- [ ] Decide if we need partition key specified for everything (needs debate)
- [ ] ???

I'd prefer to do one big breaking change PR, so need to go through our list still to see if I've missed anything. Feel free to modify the PR todo list if you see anything.


christopheranderson:
FYI - a bunch of files got touched when I was trying to remove circular dependencies that started to break my build. We should think about having rollup actually throw on new circular dependencies being added. I had to manually scan through the log output, but I got rid of them all.

src/client/Container/Container.ts
```diff
@@ -109,7 +141,7 @@ export class Container {
    * Use `.userDefinedFunctions` for creating new user defined functions, or querying/reading all user defined functions.
    * @param id The id of the {@link UserDefinedFunction}.
    */
-  public userDefinedFunction(id: string): UserDefinedFunction {
+  protected __userDefinedFunction(id: string): UserDefinedFunction {
```
southpolesteve: Why adding the double underscore prefix?

src/request/FeedOptions.ts
```diff
@@ -1,47 +1,47 @@
-import { IHeaders } from "../index";
+import { CosmosHeaders } from "../index";
 
 /**
  * The feed options and query methods.
  */
-export interface FeedOptions {
+export class FeedOptions {
```
southpolesteve: Why make this a class?

tsconfig.commonjs.json
```diff
@@ -0,0 +1,20 @@
+{
```
southpolesteve:
```suggestion
{
  "extends": "./tsconfig.json",
  "compilerOptions": {
    "module": "commonjs"
  }
}
```

src/queryIterator.ts
```diff
@@ -135,27 +125,29 @@ export class QueryIterator<T> {
   /**
    * Retrieve all the elements of the feed and pass them as an array to a function
    */
-  public async toArray(): Promise<Response<T[]>> {
+
+  public async toArray(): Promise<FeedResponse<T>> {
```
southpolesteve: Do we need a new special type for the `toArray` repsonse that remembers all the headers? I think the suffers from the same issue we have in v2 where toArray only has the last header values

src/client/Container/Container.ts
christopheranderson: single underscore is yelled at by tslint and I want a different name than the old type so we can still build our mocked types using the old api surface area.

src/queryIterator.ts
christopheranderson: This is a good point. I'll think on it.

src/request/FeedOptions.ts
christopheranderson: I think this is a mistake. I'm not sure why I did this.

src/queryIterator.ts
christopheranderson: We're not saving any headers anymore, so 1 FeedResponse is good enough.

christopheranderson:
Created #248 to track fixing up samples after these changes.

PR #240, opened 2019-02-20
Fix default indexing tests

southpolesteve:


src/test/functional/container.spec.ts
```diff
@@ -159,130 +159,116 @@ describe("NodeJS CRUD Tests", function() {
   });
 
   describe("Validate container indexing policy", function() {
-    const indexPolicyTest = async function() {
```
southpolesteve: Indenting changed because I removed unnecessary try/catch blocks

src/test/functional/container.spec.ts
```diff
@@ -292,24 +278,8 @@ describe("NodeJS CRUD Tests", function() {
 
       assert(rootIncludedPath); // root path should exist.
 
-      // In the root path, there should be one HashIndex for Strings, and one RangeIndex for Numbers.
+      // In the root path, there should be two indexes. One for Strings and one for Numbers.
       assert.equal(2, rootIncludedPath["indexes"].length);
-
-      let hashIndex: Index = null;
```
southpolesteve: This part of the tests seems overkill to me. I just removed it entirely

PR #246, opened 2019-03-05
Refactor ConnectionPolicy and RetryOptions to interface

southpolesteve:


src/documents/ConnectionPolicy.ts
```diff
@@ -1,36 +1,37 @@
...
-
+export interface ConnectionPolicy {
   /** Determines which mode to connect to Cosmos with. (Currently only supports Gateway option) */
-  public ConnectionMode = ConnectionMode.Gateway;
-  /** Attachment content (aka media) download mode. Should be one of the values of {@link MediaReadMode} */
-  public MediaReadMode: keyof typeof MediaReadMode = MediaReadMode.Buffered;
-
-  /** Time to wait for response from network peer for attachment content (aka media) operations. Represented in milliseconds. */
-  public MediaRequestTimeout = ConnectionPolicy.defaultMediaRequestTimeout;
+  ConnectionMode?: ConnectionMode;
```
southpolesteve: @christopheranderson How do you feel about making all CosmosClientOptions snakeCase? This would be a good time to do it

package.json
```diff
@@ -30,6 +30,7 @@
     "bundle": "rollup -c",
     "bundle-types": "node bundle-types.js",
     "build": "npm run clean && npm run check-format && npm run lint && npm run compile && npm run docs && npm run bundle && npm run bundle-types",
+    "test:ci": "npm test -- --forbid-only ",
```
southpolesteve: Once this get merged in to master I can switch CI to run this instead

src/documents/ConnectionPolicy.ts
southpolesteve: Resolved in person. We are switching all options to use snakeCase

package.json
christopheranderson: Could you quickly describe what this does?

southpolesteve: I keep accidentally committing a `.only` in test files. This will error if any test file contains `.only` but the idea is to only have it run that way in CI. Locally you can `.only` all you want

PR #252, opened 2019-03-08
Switch _ts type from string to number

### LINKED ISSUE #251: Incorrect type for Resource._ts
When working with resources using the `@azure/cosmos` package, a `Resource` type is returned with most APIs, which contains the common base schema. This works well, but the `_ts` property in the `Resource` type is defined as a `string`, rather than a `number`:

https://github.com/Azure/azure-cosmos-js/blob/ae53dbaaa996dbd56ba0e1b55bb95dfd7b4b510d/src/client/Resource.ts#L1-L13

From the actual data returned, as well as the [REST API examples](https://docs.microsoft.com/en-us/rest/api/cosmos-db/get-a-document#body-1), it seems like this type should be a number instead of a string.

Environment:

 - Node.js version: 8.14.0
 - @azure/cosmos version: 2.1.1

princjef:
Fixes #251.

Also removes an extra comment for a property that is no longer defined in the type.


southpolesteve:
Thank you!

PR #263, opened 2019-03-14
User Cancellation of Requests

southpolesteve:
Allow users to pass AbortController.signal in options

### DIFF
--- a/src/retry/retryUtility.ts
+++ b/src/retry/retryUtility.ts
@@ -71,7 +71,6 @@ export async function execute({
       defaultRetryPolicy: new DefaultRetryPolicy(requestContext.operationType)
     };
   }
-  const httpsRequest = executeRequest(requestContext);
   if (!requestContext.locationRouting) {
     requestContext.locationRouting = new LocationRouting();
   }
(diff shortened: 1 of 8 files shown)


src/retry/retryUtility.ts
```diff
@@ -57,40 +51,42 @@ export async function execute({
...
-        request.operationType,
+        requestContext.resourceType,
+        requestContext.operationType,
         connectionPolicy
       ),
-      defaultRetryPolicy: new DefaultRetryPolicy(request.operationType)
+      defaultRetryPolicy: new DefaultRetryPolicy(requestContext.operationType)
     };
   }
-  const httpsRequest = createRequestObjectFunc(connectionPolicy, requestOptions, body);
```
southpolesteve: @christopheranderson this change one line but significant. Before we were creating this promise up here and awaiting inside the try block on line 89. With abort controller, this caused unhandled promise errors. So I moved it down to line 89, but does that change all the location routing stuff? I could move that into the try block to

southpolesteve: Resolved this on v3 already in the RequestContext PR

PR #273, opened 2019-03-19
endpoint variable name needs to be as is

prashanthmadi:
spent an hour digging through code on why im unable to make connection only to realize that i had it as end`P`oint instead of endpoint as variable name :(. you can ignore this if you feel its silly.


southpolesteve:
Thanks, for the PR but this isn't something we are going to change.  `{ endpoint }` is valid JS shorthand for `{ endpoint: endpoint }`. I'd highly recommend checking out TypeScript which would catch a misspelled property shorthand.

PR #277, opened 2019-03-21
Allow blank string partition keys. Fixes #249

### LINKED ISSUE #249: Empty string partitionKey does not work with some methods
Example of a call that fails:

`await container.item("id", "").delete()`

If anyone runs into this, you can work around by specifying the partition key on the options object:

`await container.item("id").delete({ partitionKey: "" })`

southpolesteve:
Also fixes some old docs


test/integration/crossPartition.spec.ts
```diff
@@ -268,10 +259,12 @@ describe("Cross Partition", function() {
         maxDegreeOfParallelism: 0
       };
 
-      const expectedOrderedIds = [1, 10, 18, 2, 3, 13, 14, 16, 17, 0, 11, 12, 5, 9, 19, 4, 6, 7, 8, 15];
```
southpolesteve: @christopheranderson This PR removes order checking from the first few (non-ordered) tests here. They have broken several times. I don't think we should be verifying ordering here because it is not guaranteed by the backend.

src/request/RequestHandler.ts
```diff
@@ -38,6 +38,7 @@ export class RequestHandler {
 
     try {
       // TODO Remove any
+      // console.log((requestOptions as any).href + requestOptions.path, requestOptions.method);
```
christopheranderson: remove this

test/integration/crossPartition.spec.ts
christopheranderson: I think you're right. What we really care about is that it makes multiple backend requests to different partitions. We really need to be able to intercept the 400 response to be assured it went cross-partition. If you don't want to do that in this PR, open a TODO issue.

PR #278, opened 2019-03-21
Remove agent socket defaults

southpolesteve:
256 is low and arbitrary? We should use the node defaults https://nodejs.org/api/http.html#http_new_agent_options


tony-gutierrez:
In the normal world it's low, in the azure app service world it is double the available sockets

christopheranderson:
@tony-gutierrez - for consumption Azure Functions you're right. App Service non-consumption doesn't have the same limits as consumption plan, I believe.

Either way, we have a fair number of incidents we get that are related to having a really low limit and needing to explain to people how to raise it. If you're in Functions land, you likely are better off passing your own Agent anyway so you can reuse it with any other http clients you're using.

southpolesteve:
Added https://github.com/Azure/azure-cosmos-js/issues/281 to track how we apply this to v3. It is gonna be different

PR #280, opened 2019-03-22
Replace cross-fetch with node-fetch

southpolesteve:
cross-fetch will automatically ship a fetch polyfill to all browsers. node-fetch has better behavior which is to rely on a user-supplied polyfill in the browser.


southpolesteve:
Yep. IE 11 Users will need a polyfill like https://github.com/github/fetch. Everyone else should be fine.

PR #285, opened 2019-03-28
Remove Extra JSON.parse in Cross Partition Queries

southpolesteve:


src/queryExecutionContext/partitionedQueryExecutionContextInfoParser.ts
```diff
@@ -1,47 +1,47 @@
-/** @hidden */
-const PartitionedQueryContants = {
-  QueryInfoPath: "queryInfo",
-  TopPath: ["queryInfo", "top"],
-  OrderByPath: ["queryInfo", "orderBy"],
-  AggregatePath: ["queryInfo", "aggregates"],
-  QueryRangesPath: "queryRanges",
-  RewrittenQueryPath: ["queryInfo", "rewrittenQuery"]
-};
+// /** @hidden */
```
christopheranderson: Delete me?

src/request/ErrorResponse.ts
```diff
@@ -6,12 +6,19 @@ interface ErrorBody {
   additionalErrorInfo?: AdditionalErrorInfo;
 }
 
-interface AdditionalErrorInfo {
+export interface AdditionalErrorInfo {
```
christopheranderson: Could we maybe call this PartitionedQueryExecutionInfo or something along those lines? I know that this is the property name we use in the error body, but it is not very descriptive (as it is in the error body itself). We'll end up using this same object body once Query Plan becomes available soon-ish.

PR #286, opened 2019-03-28
Cleanup some unecessary args in retry utility

southpolesteve:
Fallout from moving everything into RetryContext.


christopheranderson:
It's sad we don't have any test coverage for this code.

PR #293, opened 2019-04-04
Add activityId to response and custom executeRequest

southpolesteve:


src/CosmosClientOptions.ts
```diff
@@ -31,4 +38,8 @@ export interface CosmosClientOptions {
    * Use an agent such as https://github.com/TooTallNate/node-proxy-agent if you need to connect to Cosmos via a proxy
    */
   agent?: Agent;
+  /** An optional function to be called on each request
+   * // TODO Add docs
+   */
+  executeRequest?: ExecuteRequest;
```
southpolesteve: @christopheranderson I'm definitely not set on this API. I would like to maybe rethink it when I do the pipeline work but for now, I need something here to get this into the portal

christopheranderson: Why do we need this executeRequest thing?

src/request/FeedResponse.ts
```diff
@@ -16,4 +16,7 @@ export class FeedResponse<TResource> {
   public get requestCharge(): number {
     return this.headers[Constants.HttpHeaders.RequestCharge];
   }
+  public get activityId(): number {
```
christopheranderson: Please something to our tests to make sure it is being set in single and cross partition scenarios.

christopheranderson: string
```suggestion
  public get activityId(): string{
```

src/CosmosClientOptions.ts
christopheranderson: Is this just needed for the portal? Could we do it in a hacky way where we expose it as a public type on ClientContext so you can override it?

PR #296, opened 2019-04-04
Output transpiled test files to dist-test directory

jeremymeng:
to keep test directory clean

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -268,8 +268,4 @@ temp/
 ts-test/package-lock.json
 ts-test/package.json
 ts-test/*.js
-test/**/*.js
-test/**/*.js.map
-test/**/*.d.ts
-test/
 tsconfig.tsbuildinfo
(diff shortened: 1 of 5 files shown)


jeremymeng:
I found that output files along side *.ts files make navigating in the VS Code explorer a little harder. thoughts? @southpolesteve

southpolesteve:
I would actually like to go the other direction and not emit at all. Both @christopheranderson and I have hit bugs with typescript projects and vscode and it has been a bit frustrating. I think my ideal setup is:

- `npm run compile` should just type check, no emit
- `npm run test` should use ts-node/transpileOnly to run .ts tests directly
-  `npm run build` should emit just src since it has some extra compiler options like noUnusedLocals

jeremymeng:
Does ts-node work with browser tests? (I noticed that npm run test-browser seems broken in 3 branch.)

Sent from my Windows 10 device

________________________________

southpolesteve:
Browser tests aren't working right now. Its something we need to fix.

jeremymeng:
Abandoning

PR #299, opened 2019-04-08
Remove test catch all for .gitignore

christopheranderson:
Have to force add tests, but the other filters catch the random .js/.js.map files.


southpolesteve:
This is prob why my search was broken

Issue #301, opened 2019-04-08
FeedResponse.activityId on cross partition queries


southpolesteve:
A cross partition request can generate many different activity ids. We need to figure out some way to track these and expose them to the caller. Right now `FeedResponse.activityId` will always be undefined as the last operation returns a fake header bag with empty query metrics and request charge.

PR #304, opened 2019-04-11
Query Plan API

southpolesteve:
- Remove proxy query execution context
- enableCrossPartitionQuery is now true by default and not a user passed option
- Add error and test for non-VALUE aggregates


src/queryIterator.ts
```diff
@@ -64,7 +67,7 @@ export class QueryIterator<T> {
     this.reset();
     let index = 0;
     while (this.queryExecutionContext.hasMoreResults()) {
-      const result = await this.queryExecutionContext.nextItem();
+      const result = await this.executeMethod("nextItem");
```
bchong95: Isn't this a bit fragile? what if someone changes the name of the method? And does it come with intellisense support?

src/queryIterator.ts
```diff
@@ -35,8 +38,8 @@ export class QueryIterator<T> {
     this.fetchFunctions = fetchFunctions;
```
bchong95: please null check where appropriate.

src/queryIterator.ts
```diff
@@ -174,13 +177,41 @@ export class QueryIterator<T> {
...
+    return (
+      error.code === StatusCodes.BadRequest &&
+      "substatus" in error &&
+      error["substatus"] === SubStatusCodes.CrossPartitionQueryNotServable
+    );
+  }
+
+  private async executeMethod(methodName: Exclude<keyof IExecutionContext, "hasMoreResults">) {
+    try {
+      return await this.queryExecutionContext[methodName]();
```
bchong95: can we do a callback function instead?

src/queryIterator.ts
```diff
@@ -174,13 +177,41 @@ export class QueryIterator<T> {
...
+    } catch (err) {
+      if (this.isPartitionedExecutionError(err)) {
+        // if this's a partitioned execution info switches the execution context
+        const partitionedExecutionInfo = err.body.additionalErrorInfo;
+        this.queryExecutionContext = this.createPipelinedExecutionContext(partitionedExecutionInfo);
+        try {
+          // retry call with same method
+          return this.queryExecutionContext[methodName]();
+        } catch (e) {
+          throw e;
```
bchong95: do we need a try catch if we are just going to throw the error?

src/queryIterator.ts
southpolesteve: It does work fully with intellisense. The type below `Exclude<keyof IExecutionContext, "hasMoreResults">` gives us pretty good checking.  I'll noodle on if there is a better way to do this.

src/queryIterator.ts
southpolesteve: Trying to avoid callbacks in favor of async/await

src/queryIterator.ts
southpolesteve: You are correct. Will remove

src/queryExecutionContext/EndpointComponent/AggregateEndpointComponent.ts
```diff
@@ -10,6 +10,7 @@ export class AggregateEndpointComponent implements ExecutionContext {
   private aggregateValues: any[];
   private aggregateValuesIndex: number;
   private localAggregators: any[];
+  private started: boolean;
```
southpolesteve: @bchong95 This is the only significant change to logic. I can't quite tell why this changed, but in the new query plan approach, this component returns false initially for `hasMoreResults`. Adding this boolean to track state fixes.

test/functional/query.spec.ts
```diff
@@ -133,27 +93,16 @@ describe("NodeJS CRUD Tests", function() {
...
-      await queryIteratorToArrayTest();
-    });
-
-    it("validate queryIterator asyncIterator", async function() {
-      await queryIteratorAsyncIteratorTest();
-    });
-
-    it("nativeApi validate queryIterator iterator executeNext name based", async function() {
-      await queryIteratorExecuteNextTest();
+      // TODO: Continuation tokens are no longer returned by fetchNext because it goes down the pipelines execution context path - stfaul
```
southpolesteve: @christopheranderson Let's discuss this tomorrow. Now that everything goes through the pipelined query execution context, we lose continuation tokens for simple queries

test/integration/query.spec.ts
```diff
@@ -64,8 +63,9 @@ describe("Test Query Metrics", function() {
 
       while (queryIterator.hasMoreResults()) {
         const { resources: results, queryMetrics, activityId, requestCharge } = await queryIterator.fetchNext();
-        assert(activityId, "activityId must exist");
-        assert(requestCharge, "requestCharge must exist");
+        // TODO: stfaul - plumb activityId and requestCharge through the iterator and renable these
```
southpolesteve: @christopheranderson We need to land your other PR or I need to pipe these values through

southpolesteve: https://github.com/Azure/azure-cosmos-js/issues/301

src/ClientContext.ts
```diff
@@ -138,6 +141,45 @@ export class ClientContext {
...
+
+    request.endpoint = await this.globalEndpointManager.resolveServiceEndpoint(request);
+    request.headers = await this.buildHeaders(request);
+    request.headers[Constants.HttpHeaders.IsQueryPlan] = "True";
+    request.headers[Constants.HttpHeaders.QueryVersion] = "1.4";
+    request.headers[Constants.HttpHeaders.SupportedQueryFeatures] =
+      "Aggregate, Distinct, MultipleOrderBy, OffsetAndLimit, OrderBy, Top, CompositeAggregate";
+    request.headers[Constants.HttpHeaders.ContentType] = Constants.MediaTypes.QueryJson;
+    if (typeof query === "string") {
+      request.body = { query }; // Converts query text to query object.
```
bchong95: I think this should be in the format:

```
{
    "query" : <query>,
    "parameters": []
}
```

Can you add one test to see that the query plan code work with parameters?

src/queryExecutionContext/EndpointComponent/AggregateEndpointComponent.ts
```diff
@@ -50,50 +51,43 @@ export class AggregateEndpointComponent implements ExecutionContext {
...
-        });
-      });
-
-      // Get the aggregated results
+    resources.forEach((resource: any) => {
+      // TODO: any
       this.localAggregators.forEach(aggregator => {
-        this.aggregateValues.push(aggregator.getResult());
+        let itemValue;
+        // Get the value of the first property if it exists
```
bchong95: Unrelated to this change:

Can we create a work item to track aggregation pipeline using the higher precision aggregate result?

The changes related to this file will explain:

https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB?path=%2FProduct%2FSDK%2F.net%2FMicrosoft.Azure.Documents.Client%2FQuery%2FAggregation%2FMinMaxAggregator.cs&version=GBmaster&_a=contents

src/queryExecutionContext/EndpointComponent/AggregateEndpointComponent.ts
bchong95: I think it's because aggregates are a unique case where you can not return a single document until the source components are fully drained (have no more results). The default behavior for `hasMoreResults` is just whether the source has more results, which is not what you want for aggregates.

src/request/RequestHandler.ts
```diff
@@ -42,6 +42,7 @@ async function httpRequest(requestContext: RequestContext) {
   }
 
   try {
+    // console.log(requestContext.method, requestContext.path);
```
bchong95: Can this be removed?

test/integration/aggregateQuery.spec.ts
```diff
@@ -43,220 +41,179 @@ describe("NodeJS Aggregate Query Tests", async function() {
...
-          partitionKey,
-          uniquePartitionKey
-        );
-        let testName = util.format("%s SinglePartition %s", config.operator, "SELECT VALUE");
-        testConfigs.push({
-          testName,
-          query,
-          expected: config.expected
-        });
+      query = `SELECT VALUE ${operator}(r.${partitionKey}) FROM r WHERE ${condition} ORDER BY r.${partitionKey}`;
```
bchong95: Can we test that aggregates without the `VALUE`. If it's v3 SDK they should be blocked. If v2 SDK they should work in a single physical partition to avoid a breaking change.

test/integration/query.spec.ts
```diff
@@ -30,8 +30,7 @@ describe("ResourceLink Trimming of leading and trailing slashes", function() {
 
     await container.items.create(doc);
     const query = "SELECT * from " + containerId;
-    const queryOptions = { partitionKey: "pk" };
-    const queryIterator = container.items.query(query, queryOptions);
+    const queryIterator = container.items.query(query);
```
bchong95: You might already have this but can you assert that unsupported queries are not allowed:

like GROUP BY

src/queryExecutionContext/EndpointComponent/AggregateEndpointComponent.ts
southpolesteve: https://github.com/Azure/azure-cosmos-js/issues/340

test/integration/aggregateQuery.spec.ts
southpolesteve: Yep. I have a test for this at the bottom of the file. They are blocked.

test/integration/query.spec.ts
southpolesteve: This works now that we attempt to hit the gateway with the query first.

src/ClientContext.ts
southpolesteve: We test this in some of the cross partition specs already

src/request/RequestHandler.ts
southpolesteve: Yep!

test/integration/query.spec.ts
southpolesteve: I added a test for GROUP BY erroring in the aggregates file

PR #306, opened 2019-04-13
Add support for LIMIT and OFFSET in queries

southpolesteve:


src/queryExecutionContext/EndpointComponent/IEndpointComponent.ts
```diff
@@ -1,9 +0,0 @@
-import { Response } from "../../request";
```
southpolesteve: This was the same as `ExecutionContext` so removed it

src/queryExecutionContext/EndpointComponent/OffsetLimitEndpointComponent.ts
```diff
@@ -1,53 +1,49 @@
...
    * @constructor TopEndpointComponent
    * @param { object } executionContext - Underlying Execution Context
    * @ignore
    */
-  constructor(private executionContext: IExecutionContext, private topCount: number) {}
+  constructor(private executionContext: ExecutionContext, private limit: number, private offset: number) {}
 
   /**
    * Execute a provided function on the next element in the TopEndpointComponent.
    * @memberof TopEndpointComponent
```
bchong95: Nit: This comment needs to be updated.

src/queryExecutionContext/EndpointComponent/OffsetLimitEndpointComponent.ts
```diff
@@ -1,53 +1,49 @@
...
-export class TopEndpointComponent implements IEndpointComponent {
+export class OffsetLimitEndpointComponent implements ExecutionContext {
   /**
    * Represents an endpoint in handling top query. It only returns as many results as top arg specified.
    * @constructor TopEndpointComponent
    * @param { object } executionContext - Underlying Execution Context
    * @ignore
    */
-  constructor(private executionContext: IExecutionContext, private topCount: number) {}
+  constructor(private executionContext: ExecutionContext, private limit: number, private offset: number) {}
```
bchong95: Can we switch the argument order, so that it's offset, then limit (to match the query operation).

src/queryExecutionContext/pipelinedQueryExecutionContext.ts
```diff
@@ -63,10 +60,17 @@ export class PipelinedQueryExecutionContext implements IExecutionContext {
       this.endpoint = new AggregateEndpointComponent(this.endpoint, aggregates);
     }
 
-    // If top then add that to the pipeline
+    // If top then add that to the pipeline. TOP N is effectively LIMIT N OFFSET 0
     const top = partitionedQueryExecutionInfo.queryInfo.top;
     if (typeof top === "number") {
-      this.endpoint = new TopEndpointComponent(this.endpoint, top);
+      this.endpoint = new OffsetLimitEndpointComponent(this.endpoint, top, 0);
+    }
```
bchong95: Do we need error handling cases for these?

test/integration/crossPartition.spec.ts
```diff
@@ -867,6 +724,103 @@ describe("Cross Partition", function() {
...
+      // validates the results size and order
+      await executeQueryAndValidateResults({
+        query: querySpec,
+        options,
+        expectedOrderIds: expectedOrderedIds
+      });
+    });
+
+    it("Validate OrderBy Query With ASC and LIMIT 0 and OFFSET 5", async function() {
+      const limit = 5;
```
bchong95: Since you can pro grammatically generate the expected value, can you use nested loops for different combinations of offset, limit, and maxItemCount?

test/integration/crossPartition.spec.ts
```diff
@@ -867,6 +724,103 @@ describe("Cross Partition", function() {
       }
     });
 
+    it("Validate simple LIMIT OFFSET", async function() {
```
bchong95: Can we add one test for an OFFSET LIMIT query with a logical partition key?

src/queryExecutionContext/pipelinedQueryExecutionContext.ts
southpolesteve: I don't think so. If we're getting back an error happy to bubble up to the user

test/integration/crossPartition.spec.ts
southpolesteve: Prefer not to at the moment. The entire file needs work, but for now would like to keep it consistent with the other test.

test/integration/crossPartition.spec.ts
southpolesteve: There is a larger refactor that needs to happen to support this test. Will add a TODO for it.

christopheranderson:
We should open a bug that tracks adding the query plan endpoint swap before we can GA this.

PR #307, opened 2019-04-22
Add DISTINCT query support

southpolesteve:
Should be merged after #306


src/queryExecutionContext/EndpointComponent/OrderedDistinctEndpointComponent.ts
```diff
@@ -0,0 +1,31 @@
+import { sha1 } from "crypto-hash";
```
southpolesteve: @bchong95 @christopheranderson I ended up going with sha1 (160 bits) for this hash as it is basically zero bundle cost. The browser and node both have a built-in sha1 hash and the `crypto-hash` package handles loading the correct one in each environment. Happy to switch this over to murmurhash once that returns, but this seems like the simplest thing for now.

bchong95: Im good with sha1, since the hash doesn't need to be crypographically secure:

https://en.wikipedia.org/wiki/SHA-1#SHAttered_%E2%80%93_first_public_collision

Only vulnerablity I can think of is if someone using CosmosDB allows a user to insert user data and they generate a collision to make another person's document not show up in the distinct result. But as the wiki says it would like 1100 years of CPU time, which I am not too worried about.

southpolesteve: I'll put something on the calendar for 3119

src/queryExecutionContext/EndpointComponent/UnorderedDistinctEndpointComponent.ts
```diff
@@ -0,0 +1,33 @@
+import { sha1 } from "crypto-hash";
+import stableStringify from "fast-json-stable-stringify";
+import { Response } from "../../request";
+import { ExecutionContext } from "../ExecutionContext";
+
+/** @hidden */
+export class UnorderedDistinctEndpointComponent implements ExecutionContext {
```
bchong95: Can we make these classes implement some `DistinctEndpointComponentBase` class and make these private nested. The base class can then have a factory create method that take the `DistinctType` and `new`s up the appropriate ExecutionContext.

src/queryExecutionContext/pipelinedQueryExecutionContext.ts
```diff
@@ -72,6 +74,15 @@ export class PipelinedQueryExecutionContext implements ExecutionContext {
     if (typeof limit === "number" && typeof offset === "number") {
       this.endpoint = new OffsetLimitEndpointComponent(this.endpoint, offset, limit);
     }
+
+    // If distinct then add that to the pipeline
+    const distinctType = partitionedQueryExecutionInfo.queryInfo.distinctType;
```
bchong95: Can we move this to the factory method of a distinct document query execution component base class. Ordered vs unordered is an implementation detail for `DISTINCT` and the pipeline should not really concern itself with it. If the distinct component wanted to it can new up an unordered componenet for every query if it wanted to.

test/integration/crossPartition.spec.ts
```diff
@@ -340,6 +341,34 @@ describe("Cross Partition", function() {
       await executeQueryAndValidateResults({ query, options, expectedOrderIds: expectedOrderedIds });
     });
 
+    it("Validate DISTINCT Query", async function() {
+      // simple order by query in string format
+      const query = "SELECT DISTINCT VALUE r.spam3 FROM root r";
+      const options = {
+        enableCrossPartitionQuery: true,
+        maxItemCount: 2
```
bchong95: Please at parallelism to test any possible race conditions.

test/integration/crossPartition.spec.ts
```diff
@@ -340,6 +341,34 @@ describe("Cross Partition", function() {
...
+      };
+
+      const expectedOrderedIds = Array.from(new Set(documentDefinitions.map(r => r["spam3"])));
```
bchong95: Is `Set` ordered? If not then the order is undefined and this test will pass by chance.

test/integration/crossPartition.spec.ts
```diff
@@ -340,6 +341,34 @@ describe("Cross Partition", function() {
...
+
+    it("Validate DISTINCT OrderBy Query", async function() {
+      // simple order by query in string format
+      const query = "SELECT DISTINCT VALUE r.spam3 FROM root r order by r.spam3";
+      const options = {
+        enableCrossPartitionQuery: true,
+        maxItemCount: 2
+      };
+
+      const expectedOrderedIds = Array.from(new Set(documentDefinitions.map(r => r["spam3"])));
```
bchong95: The test code seems the same here as the unordered variant. Shouldn't the results differ?

test/integration/crossPartition.spec.ts
```diff
@@ -340,6 +341,34 @@ describe("Cross Partition", function() {
...
+      // validates the results size and order
+      await executeQueryAndValidateResults({ query, options, expectedOrderIds: expectedOrderedIds });
+    });
+
+    it("Validate DISTINCT OrderBy Query", async function() {
+      // simple order by query in string format
+      const query = "SELECT DISTINCT VALUE r.spam3 FROM root r order by r.spam3";
+      const options = {
+        enableCrossPartitionQuery: true,
+        maxItemCount: 2
```
bchong95: Please add tests that vary the `maxItemCount`. `maxItemCount` = 1 and collectionSize are special edge cases.

`maxItemCount` = 1 means the backend will essentially do no work and client is responsible for doing all the DISTINCT
`maxItemCount` = collectionSize means the backend will essential do all the work and the clients job is just not to lose any documents.

test/integration/crossPartition.spec.ts
southpolesteve: Yep! It was actually flaky. Sometimes they matched. Fixed

test/integration/crossPartition.spec.ts
southpolesteve: Done

test/integration/crossPartition.spec.ts
southpolesteve: Removed `Set` here

src/queryExecutionContext/pipelinedQueryExecutionContext.ts
southpolesteve: Prefer to keep this explicit for now. Trying to come back and refactor the query pipeline later and would like to have less inheritance and complexity

src/queryExecutionContext/EndpointComponent/UnorderedDistinctEndpointComponent.ts
southpolesteve: Same as other comments, prefer to keep this explicit for now. Trying to come back and refactor the query pipeline later and would like to have less inheritance and complexity

Issue #261, opened 2019-03-13
[v3] Partition Key Changes


christopheranderson:
v3 Partition Key related behaviors
1. SDK should use new API version "2018-12-31"
    - new API version blocks creation of single partition containers and makes existing ones appear to be partitioned (with everything in the default partition key `EMPTY_PARTITION_KEY`
2. SDK should transparently handle detecting system defined partition key and pass the correct `EMPTY_PARTITION_KEY` header. `"[{}]"` should be used for non-migrated containers. `[]` should be used for migrated containers. 
3. SDK should allow users to pass the partition key value explicitly
    - note, that it still has to match what is in the body that is sent to the backend
4. SDK should implicitly infer non-undefined/empty partition keys if it is not specified explicitly
    - Note, users must specify the partition key explicitly if it is empty/undefined. This is intended to prevent errors. Open for feedback, though.
5. SDK should always send a partition key definition to the backend on container create. if one is not specified, a default "_partitionKey" value will be used.
6. SDK should export a `DEFAULT_PARTITION_KEY` constant which can be used to indicate empty or undefined.

southpolesteve:
Closed by https://github.com/Azure/azure-cosmos-js/pull/308

PR #308, opened 2019-04-29
Partition Key Changes

southpolesteve:
This PR:
- Updates API version to 2018-12-31 (NP->P)
- Makes partition key required for item getter
- Provides a default partition key definition if not provided when creating a container
- Handles empty vs undefined partition key header value
- Removes Hash IndexKind support (due to API version bump)
- Removes upsert support for scripts (not supported in partitioned containers)


src/client/Item/Item.ts
```diff
@@ -32,15 +32,6 @@ export class Item {
...
 
-  /**
-   * Read the item's definition.
-   *
-   * There is no set schema for JSON items. They may contain any number of custom properties.
-   *
-   * @param options Additional options for the request, such as the partition key.
-   * Note, if you provide a partition key on the options object, it will override the primary key on `this.partitionKey`.
-   */
-  public read(options?: RequestOptions): Promise<ItemResponse<ItemDefinition>>;
```
christopheranderson: Why you delete the non-T version of this method? I'm open to that change, but seems like it's increasing the scope of what this PR is trying to do.

src/client/Item/Item.ts
```diff
@@ -66,11 +57,15 @@ export class Item {
    * ({body: item} = await item.read<TodoItem>());
    * ```
    */
-  public read<T extends ItemDefinition>(options?: RequestOptions): Promise<ItemResponse<T>>;
-  public async read<T extends ItemDefinition>(options?: RequestOptions): Promise<ItemResponse<T>> {
+  public async read<T extends ItemDefinition = any>(options: RequestOptions = {}): Promise<ItemResponse<T>> {
     options = options || {};
-    if ((!options || !options.partitionKey) && this.partitionKey !== undefined) {
-      options.partitionKey = this.partitionKey;
+    if (!options.partitionKey) {
```
christopheranderson: Did we want to prevent folks from passing partition key in request options now?

christopheranderson: This applies in a few places, but I'm not gonna repeat my comment for them.

src/client/Item/Item.ts
```diff
@@ -135,12 +124,16 @@ export class Item {
...
-    options = options || {};
-    if ((!options || !options.partitionKey) && this.partitionKey !== undefined) {
-      options.partitionKey = this.partitionKey;
+  public async delete<T extends ItemDefinition = any>(options: RequestOptions = {}): Promise<ItemResponse<T>> {
+    if (!options.partitionKey) {
+      if (this.partitionKey !== undefined) {
+        options.partitionKey = this.partitionKey;
+      } else {
+        const { resource: partitionKeyDefinition } = await this.container.getPartitionKeyDefinition();
+        options.partitionKey = extractPartitionKey(undefined, partitionKeyDefinition);
```
christopheranderson: It seems really weird to pass an undefined body to this. Shouldn't this always be this.partitionKey? Couldn't you replace this whole if/else block with something like:

```typescript
options.partitionKey = this.partitionKey !== undefined 
                                      ? this.partitionKey 
                                      : (partitionKeyDefintion.IsSystem ? [] : [{}]);
```

src/common/partitionKeys.ts
```diff
@@ -0,0 +1,12 @@
...
+ * @hidden
+ * @internal
+ */
+export const DEFAULT_PARTITION_KEY_PATH = "/_partitionKey";
+/**
+ * Header value used for indicating undefined partition key value
+ * @hidden
+ * @internal
+ */
+export const EMPTY_PARTITION_KEY = [{}] as [{}];
```
christopheranderson: EMPTY is known as `[]` and UNDEFINED is known as `[{}]` by the internal team

The name this is given in all the other SDKs is `NONE_PARTITION_KEY_VALUE`, but I'd imagine we should just handle undefined instead and not have a well known constant we use.

test/common/TestHelpers.ts
```diff
@@ -79,25 +84,31 @@ export async function bulkInsertItems(
...
+export async function bulkReadItems(container: Container, documents: any[], partitionKeyProperty: string) {
   return await Promise.all(
     documents.map(async document => {
-      const options =
-        partitionKey && document.hasOwnProperty(partitionKey)
-          ? { partitionKey: document[partitionKey] }
-          : { partitionKey: {} };
+      const partitionKey = document.hasOwnProperty(partitionKeyProperty)
+        ? document[partitionKeyProperty]
+        : EMPTY_PARTITION_KEY;
```
christopheranderson: If this is using the `[{}]` value under the covers, then it would fail on a migrated container you ran the tests on.

test/functional/authorization.spec.ts
```diff
@@ -160,7 +160,7 @@ describe("NodeJS CRUD Tests", function() {
       const { resource: successDoc } = await col1Client
         .database(entities.database.id)
         .container(entities.coll1.id)
-        .item(entities.doc1.id)
+        .item(entities.doc1.id, EMPTY_PARTITION_KEY)
```
christopheranderson: let's just use `undefined` instead of this random constant?

christopheranderson: For all cases in our tests/samples

test/functional/container.spec.ts
```diff
@@ -5,166 +5,134 @@ import { ContainerRequest } from "../../dist-esm/client/Container/ContainerReque
...
+    it("Default partition key", async function() {
+      await containerCRUDTest();
+    });
+
+    it("Custom partition key", async function() {
+      await containerCRUDTest("/id");
     });
 
-    it("nativeApi container with bad partition key definition name based", async function() {
+    it("Bad partition key definition", async function() {
```
christopheranderson: Not sure if we need a test to check that hash isn't allowed anymore. This kind of testing is prone to breaking.

test/functional/container.spec.ts
```diff
@@ -275,141 +231,113 @@ describe("NodeJS CRUD Tests", function() {
...
+      assert.notEqual(headers2[Constants.HttpHeaders.IndexTransformationProgress], undefined);
+      assert.notEqual(headers2[Constants.HttpHeaders.LazyIndexingProgress], undefined);
+
+      const noneContainerDefinition = {
+        id: "none_coll",
+        indexingPolicy: { indexingMode: IndexingMode.none, automatic: false }
+      };
+      const { headers: headers3 } = await createThenReadcontainer(database, noneContainerDefinition);
+      assert.notEqual(headers3[Constants.HttpHeaders.IndexTransformationProgress], undefined);
+      assert.equal(headers3[Constants.HttpHeaders.LazyIndexingProgress], undefined);
```
christopheranderson: we should probably move these to the container response (doesn't need to be in this PR, though)

test/functional/npcontainer.spec.ts
```diff
@@ -0,0 +1,98 @@
+import assert from "assert";
+import { CosmosClient, Constants, Container } from "../../dist-esm";
+import { removeAllDatabases, getTestContainer } from "../common/TestHelpers";
+import { endpoint, masterKey } from "../common/_testConfig";
+import { ResourceType, HTTPMethod } from "../../dist-esm/common";
+
+const legacyClient = new CosmosClient({
+  endpoint,
+  key: masterKey,
+  plugins: [
```
christopheranderson: Nice. I love it. Plugins already paying off.

test/functional/npcontainer.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    {
+      on: "request",
+      plugin: (context, next) => {
+        // Intercepts the API request to create a non-partitioned container using an old API version
+        if (context.resourceType === ResourceType.container && context.method === HTTPMethod.post) {
+          context.body = JSON.stringify({ id: JSON.parse(context.body).id });
```
christopheranderson: what is this doing? Dropping everything except the name of the container? To avoid weird differences in the API versions?

test/integration/query.spec.ts
```diff
@@ -58,9 +58,8 @@ describe("Test Query Metrics On Single Partition Collection", function() {
       const createdContainer = database.container(createdCollectionDef.id);
 
       await createdContainer.items.create(document);
-      const collectionLink = "/dbs/" + database.id + "/colls/" + collectionId + "/";
       const query = "SELECT * from " + collectionId;
-      const queryOptions: FeedOptions = { populateQueryMetrics: true };
+      const queryOptions: FeedOptions = { populateQueryMetrics: true, enableCrossPartitionQuery: true };
```
christopheranderson: Let's talk about enableCrossPartitionQuery experience later

src/client/Item/Item.ts
southpolesteve: I've been removing these and adding default types as I go. I think they can all use default types, so decreases future maintenance load.

test/functional/authorization.spec.ts
southpolesteve: Agreed

test/functional/npcontainer.spec.ts
southpolesteve: Yes. We now default to `_partitionKey` inside the SDK, so need to "undo" that for this call. IMO it would be nice if the backend accepted just an `id` and defaulted `_partitionKey` there

test/common/TestHelpers.ts
southpolesteve: Changed to undefined

src/common/partitionKeys.ts
southpolesteve: Removed entirely

src/client/Item/Item.ts
southpolesteve: Nested ternaries? You are a madman

But yes.

src/client/Item/Item.ts
southpolesteve: I'm for it. Will do.

southpolesteve: I'm gonna make these annon objects now that they differ from `RequestOptions`

test/functional/container.spec.ts
southpolesteve: https://github.com/Azure/azure-cosmos-js/issues/319

src/client/Item/Item.ts
southpolesteve: Discussed on teams and going to leave this in two places for now. Will remove in a later PR https://github.com/Azure/azure-cosmos-js/issues/320

srascar:
Hello, 
in https://github.com/Azure/azure-cosmos-js/issues/261, it is specified that the library should export a constant to support containers with no partition key: `DEFAULT_PARTITION_KEY`.

I could not find it.
Also it would be nice to export `DEFAULT_PARTITION_KEY_PATH`.

Thanks in advance.

southpolesteve:
@srascar exporting a constant ended up not being necessary. If you pass `undefined` the SDK will send the correct partition key value to the backend. I'll open a PR to export the default partition key path

srascar:
@southpolesteve Thank you very much. 
I found the `undefined` trick but I don't think I saw any documentation about it.
I mentioned it here https://github.com/srascar/galaxy-migrations#update-a-document-when-your-container-doesnt-have-partition-key

PR #309, opened 2019-04-29
Add quick reference to install syntax

craigshoemaker:
Readers can infer the install command from the first line of the sample script, but this makes it easy to find as you are scanning the README.


southpolesteve:
👍Thanks!

Issue #320, opened 2019-05-21
Remove `options.partitionKey` for item calls


southpolesteve:
There are several places where it is required as an argument, but also passable as an option. This is confusing

southpolesteve:
Completed in #322

PR #322, opened 2019-05-22
Remove partitionKey from RequestOptions

southpolesteve:
#320

### DIFF
--- a/src/client/Item/Item.ts
+++ b/src/client/Item/Item.ts
@@ -13,6 +13,7 @@ import { ItemResponse } from "./ItemResponse";
  * @see {@link Items} for operations on all items; see `container.items`.
  */
 export class Item {
+  private partitionKey: PartitionKey;
   /**
    * Returns a reference URL to the resource. Used for linking in Permissions.
    */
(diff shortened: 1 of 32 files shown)


src/client/Item/Item.ts
```diff
@@ -28,7 +29,7 @@ export class Item {
   constructor(
     public readonly container: Container,
     public readonly id: string,
-    public readonly partitionKey: string,
+    public partitionKey: PartitionKey,
```
christopheranderson: We shouldn't make this public settable. We should use a _partitionKey or partitionKeyInternal with a getter instead.

southpolesteve: I made it private and it only gets public set in the constructor

BernhardHofmann:
The [documentation for create](https://azure.github.io/azure-cosmos-js/classes/items.html#create) still states that the `RequestOptions` are the way to specify the partition key on a create. I can't find from issue #320 or this PR where or how I specify the partition key when creating an item now. 😞

I've submitted an issue for the documentation here: https://github.com/MicrosoftDocs/azure-docs/issues/83600

PR #327, opened 2019-05-30
Update to Typescript 3.5

southpolesteve:


src/client/Item/ItemDefinition.ts
```diff
@@ -7,5 +7,6 @@
  * in {@link ItemBody}
  */
 export interface ItemDefinition {
```
christopheranderson: non-blocking comment: should we even have this object? It's almost useless.

southpolesteve: I don't like it either

PR #328, opened 2019-05-31
Remove node builtin usage in favor of universal packages/APIs

southpolesteve:
- Adds universal-user-agent
- Switches usage of `url` package to global `URL`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -265,9 +265,9 @@ dist-esm/
 dist-test/
 dist-commonjs/
 temp/
-ts-test/package-lock.json
-ts-test/package.json
-ts-test/*.js
+consumer-test/package-lock.json
+consumer-test/package.json
+consumer-test/*.js
 test/**/*.js
 test/**/*.js.map
 test/**/*.d.ts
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -265,9 +265,9 @@ dist-esm/
 dist-test/
 dist-commonjs/
 temp/
-ts-test/package-lock.json
-ts-test/package.json
-ts-test/*.js
+consumer-test/package-lock.json
+consumer-test/package.json
+consumer-test/*.js
```
christopheranderson: this is a good rename

test/unit/platform.spec.ts
```diff
@@ -1,44 +1,8 @@
 import assert from "assert";
-import * as os from "os";
-import * as util from "util";
-import { getSafeUserAgentSegmentInfo, getUserAgent } from "../../dist-esm/common/index";
 import { Constants } from "../../dist-esm/index";
 
 const packageJson = require("../../package.json");
 
-describe("getUserAgent", function() {
```
christopheranderson: probably over zealous to remove it entirely, right? Should check the method still works and returns the right versions in it, right?

southpolesteve: Yep. Added test for node version and SDK

PR #330, opened 2019-06-01
Remove Buffer

### LINKED ISSUE #329: Remove usage of `Buffer`
Throws errors in the browser

southpolesteve:
Close #329

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
   },
   "dependencies": {
     "@azure/cosmos-sign": "1.0.2",
+    "atob": "2.1.2",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -79,6 +79,7 @@
   },
   "dependencies": {
     "@azure/cosmos-sign": "1.0.2",
+    "atob": "2.1.2",
```
christopheranderson: atob is dual licensed MIT and Apache 2.0. It appears to have a large number of dependencies and good deployment cadence, so not too risky. Approved ✅

PR #334, opened 2019-06-05
Throw Error subclass instead of objects for failed Requests

southpolesteve:


southpolesteve:
@christopheranderson This is a baby step toward better errors, but at least now we get stack traces.

PR #343, opened 2019-06-10
Return undefined for item reads

southpolesteve:


test/integration/incrementalFeed.spec.ts
```diff
@@ -34,9 +34,9 @@ describe("Change Feed Iterator", function() {
       await container.items.create({ id: "item2", key: "1" });
     });
 
-    after(async function() {
-      await container.delete();
-    });
+    // after(async function() {
```
christopheranderson: If this intentional, we should just delete this.

PR #346, opened 2019-06-12
Expand SQL query parameter type

### LINKED ISSUE #233: Incorrect SqlParameter type
The `SqlParameter` type restricts the `value` property to `string | number | boolean`:

https://github.com/Azure/azure-cosmos-js/blob/fbe56ef09f12111381bf7701ffcb4004cea7dad6/src/queryExecutionContext/SqlQuerySpec.ts#L23-L31

However, according to the [docs](https://docs.microsoft.com/en-us/azure/cosmos-db/how-to-sql-query#parameterized-sql-queries):

> Parameter values can be any valid JSON (strings, numbers, Booleans, null, even arrays or nested JSON). Also since Cosmos DB is schema-less, parameters are not validated against any type.

I have also observed that if I bypass Typescript's type check, array and object types are accepted at runtime by the database. Therefore, it seems like the type needs to be updated.

The most correct type for the `value` property would probably be the `JSONValue` type as defined below:

```ts
export type JSONValue = string | number | boolean | null | JSONObject | JSONArray;

export interface JSONObject {
    [x: string]: JSONValue;
}

export interface JSONArray extends Array<JSONValue> { }
```

That said, I think that it would also be fine to mark the type as either `any` or a looser form of the `JSONValue` type defined as such:

```ts
export type JSONValue = string | number | boolean | null | object | any[];
```

Environment:
 * Node.js version: 8.14.0
 * `@azure/cosmos` version: 2.1.1

southpolesteve:
Close #233

### DIFF
--- a/src/queryExecutionContext/SqlQuerySpec.ts
+++ b/src/queryExecutionContext/SqlQuerySpec.ts
@@ -27,5 +27,11 @@ export interface SqlParameter {
   /** Name of the parameter. (i.e. "@lastName") */
   name: string;
   /** Value of the parameter (this is safe to come from users, assuming they are authorized) */
-  value: string | number | boolean;
+  value: JSONValue;
 }
+
+export type JSONValue = boolean | number | string | null | JSONArray | JSONObject;
+export interface JSONObject {
+  [key: string]: JSONValue;
+}
+export interface JSONArray extends Array<JSONValue> {}


src/queryExecutionContext/SqlQuerySpec.ts
```diff
@@ -27,5 +27,11 @@ export interface SqlParameter {
   /** Name of the parameter. (i.e. "@lastName") */
   name: string;
   /** Value of the parameter (this is safe to come from users, assuming they are authorized) */
-  value: string | number | boolean;
+  value: JSONValue;
 }
+
+export type JSONValue = boolean | number | string | null | JSONArray | JSONObject;
```
christopheranderson: nit: Seems weird this type is declared before types it uses

southpolesteve: These three types are circular

PR #347, opened 2019-06-12
Enable keepAlive by default on v3

### LINKED ISSUE #281: Enable default keepAlive on v3
On v3 we no longer have a custom default agent IE https://github.com/Azure/azure-cosmos-js/pull/278. But we should figure out some way of having keepAlive on by default combined with fetch.

southpolesteve:
Close #281

### DIFF
--- a/src/request/RequestHandler.ts
+++ b/src/request/RequestHandler.ts
@@ -1,31 +1,15 @@
 import AbortController from "node-abort-controller";
 import fetch, { RequestInit, Response } from "node-fetch";
-import { isNode, trimSlashes } from "../common";
+import { trimSlashes } from "../common";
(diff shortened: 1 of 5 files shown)


src/request/RequestHandler.ts
```diff
@@ -10,6 +11,22 @@ import { RequestContext } from "./RequestContext";
 import { Response as CosmosResponse } from "./Response";
 import { TimeoutError } from "./TimeoutError";
 
+let defaultHttpAgent: Agent;
+let defaultHttpsAgent: Agent;
+
+if (isNode) {
```
christopheranderson: notion: it might be nice to just rip this into its own file since it doesn't have much to do with RequestHandler

PR #348, opened 2019-06-12
Hide plugins using internal flag

### LINKED ISSUE #325: Hide Plugins
We'll leave the code as we need it in the portal, but we're gonna make the API private for now while we work with central SDK team to sort out their pipelines work

southpolesteve:
Close #325


test/functional/npcontainer.spec.ts
```diff
@@ -19,8 +19,8 @@ const legacyClient = new CosmosClient({
         return next(context);
       }
     }
-  ]
-});
+  ] as PluginConfig[]
```
christopheranderson: this doesn't seem ideal. Couldn't we have new'd up the config on a separate line, then asset it to any and add the plugins?

```typescript
const options: CosmosClientOptions = {  /* stuff */ };
const plugins: PluginConfig = { /* stuff */ };
new CosmosClient({...options, plugins} as any);
```

That way we'd still get type help on the other settings in CosmosClientOptions, right?

southpolesteve: Yes. That is better

PR #350, opened 2019-06-17
Parse connection string for options object

southpolesteve:


test/unit/helper.spec.ts
```diff
@@ -1,15 +1,24 @@
...
       const err = {};
       const result = isResourceValid({ id: 1 }, err);
 
       assert.equal(result, false);
       assert.deepEqual(err, { message: "Id must be a string." });
-      done();
+    });
+  });
+  describe("parseConnectionString", function() {
+    it("parses a valid connection string", function() {
```
christopheranderson: Please add case for how the error message looks/etc. when it is an invalid connection string.

src/CosmosClient.ts
```diff
@@ -51,7 +52,11 @@ export class CosmosClient {
    */
 
   private clientContext: ClientContext;
-  constructor(private options: CosmosClientOptions) {
+  constructor(options: string | CosmosClientOptions) {
```
christopheranderson: I know it's annoying, but can we do this with an overload instead?

```typescript
constructor(connectionString: string)
constructor(options: CosmosClientOptions)
constructor(optionsOrConnectionString: string | CosmosClientOptions)
```

Even if you don't want to overload it, you should at least use the optionsOrConnectionString name (but I like overloading :))

christopheranderson: Also, need to update the JSdocs

test/functional/client.spec.ts
```diff
@@ -31,6 +31,10 @@ describe("NodeJS CRUD Tests", function() {
       });
       assert.ok(client !== undefined, "client shouldn't be undefined if it succeeded");
     });
+    it("Accepts a connection string", function() {
+      const client = new CosmosClient(`AccountEndpoint=${endpoint};AccountKey=${masterKey};`);
```
christopheranderson: error case

src/CosmosClient.ts
southpolesteve: Done. Added overload and docs

PR #355, opened 2019-06-20
Replace custom code with uuid library

southpolesteve:


southpolesteve:
There was a TODO item for it in code and I agreed

PR #360, opened 2019-06-25
Fix missing RUs on paginated results

southpolesteve:


christopheranderson:
Can we add a test that checks that RUs is in the expected value range for the queries where it is currently broken?

southpolesteve:
@christopheranderson done. I added it to the maxparallel = 1 test as I've found that to be the most consistent for RUs

PR #373, opened 2019-07-17
Prevent webpack from resolving modules called with `require`

southpolesteve:
This fixes 3.0.2 which is currently broken for webpack users

1. In nodejs envs, we need to set a default agent with keepAlive: true. It results in a ~2x perf improvement
2. The only way to do this with node API is to dynamically require `http` or `https` and construct the agent
3. This doesn't matter for browser users, but we need to prevent them from loading the code and node builtin modules
4. By default, webpack will see `require` and attempt to resolve the module
5. We can trick webpack into not doing this with `eval('require')`

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "main": "./dist/index.js",
   "module": "./dist-esm/index.js",
   "browser": {
-    "dist-esm/request/defaultAgent.js": "dist-esm/request/defaultAgent.browser.js"
+    "./dist-esm/request/defaultAgent.js": "./dist-esm/request/defaultAgent.browser.js"
   },
   "types": "./dist/index.d.ts",
   "engine": {
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "main": "./dist/index.js",
   "module": "./dist-esm/index.js",
   "browser": {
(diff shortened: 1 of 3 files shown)


southpolesteve:
Looks like eval is the only way to do this https://github.com/webpack/webpack/issues/8826

Tried:
- module.require. Not the same as require
- global.require. doesn't exist in node. some of the time 😜
- `typeof __non_webpack_require__ === "function" ? __non_webpack_require__ :  require`. Some how webpack is able to figure this truck out

package.json
```diff
@@ -14,7 +14,10 @@
   "version": "3.0.2",
   "author": "Microsoft Corporation",
   "main": "./dist/index.js",
-  "module": "./dist-esm/src/index.js",
+  "module": "./dist-esm/index.js",
+  "browser": {
+    "dist-esm/request/defaultAgent.ts": "dist-esm/request/defaultAgent.browser.ts"
```
bterlson: You want ".js" here.

southpolesteve:
Final update: @bterlson suggested to use the package.json browser field to accomplish this and it works great

christopheranderson:
This either needs to get merged then added to the master PR, or this should target master once that PR closes.

PR #379, opened 2019-07-19
Fixes bug when user supplies partial retry options. Close #377

### LINKED ISSUE #377: CosmosClientOptions.connectionPolicy: Nested object retryOptions replaced instead of merged
Currently if one is providing a custom set of retryOptions the default will get lost due the following LOC:
https://github.com/Azure/azure-cosmos-js/blob/v3.0.3/src/CosmosClient.ts#L64

Expected: Nested merge logic instead of merely top level merge

Solution: Nested object assign like
https://www.npmjs.com/package/nested-object-assign

southpolesteve:


southpolesteve:
https://github.com/Azure/azure-cosmos-js/issues/381

PR #382, opened 2019-07-22
Use package.json#files to prevent extraneous files from being published to npm

southpolesteve:


package.json
```diff
@@ -18,6 +18,10 @@
   "browser": {
     "./dist-esm/request/defaultAgent.js": "./dist-esm/request/defaultAgent.browser.js"
   },
+  "files": [
+    "dist",
+    "dist-esm"
```
christopheranderson: shouldn't we be distributing our license as well?

southpolesteve: LICENSE automatically gets picked up by npm no matter what you configure

PR #397, opened 2019-08-02
Add bufferItems support & other cross-partition perf improvements

christopheranderson:
Goal:

Improve performance of cross partition queries by buffering additional pages (optional) and supporting parallelism for select * queries that wouldn't otherwise force parallelism.

Changes:
- Added basic debug logging (this is not meant to replace operation level logging like the Track 2 effort wants, more for debugging/etc.)
- Adds prefetching support if bufferItems in feedOptions is enabled (In the future this can be enabled by default, but want to avoid RU spikes)
- Adds check if maxDegreeOfParallelism is set, fetch query plan and use parallelQueryExecutionContext, don't depend on 400 response.
- Fixes maxDegreeOfParallelism logic around parallelismSemaphore (mdop of 1 resulted in serial requests, 0 resulted in equivalent behavior to -1)
- Adds query benchmark sample to compare numbers
- Adds logic which forces relatively even draining of partitions for parallel non-order by queries, which reduces impact of IO bound workloads. This will change the order results get returned in because we no longer go left to right. While this is a behavior change, I don't consider it breaking.
- Fixes minAggregator/OrderBy comparators which broke when I added non-determinism to the tests 

More details:

(These numbers are on my home wifi, I should rerun them on Azure before we brag about them too much)

Random perf run with bufferItems off:
```
Scenario starting: Select * from c order by c._ts
Scenario complete: "Select * from c order by c._ts" - took 82.29s
Scenario starting: Select * from c
Scenario complete: "Select * from c" - took 101.044s
Scenario starting: Select value count(c.id) from c
Scenario complete: "Select value count(c.id) from c" - took 1.938s
```

Random perf run with bufferItems on:
```
Scenario starting: Select * from c order by c._ts
Scenario complete: "Select * from c order by c._ts" - took 57.927s
Scenario starting: Select * from c
Scenario complete: "Select * from c" - took 36.244s
Scenario starting: Select value count(c.id) from c
Scenario complete: "Select value count(c.id) from c" - took 2.648s
```

### DIFF
--- a/src/ClientContext.ts
+++ b/src/ClientContext.ts
@@ -141,7 +141,10 @@ export class ClientContext {
       }
     }
     this.applySessionToken(request);
-    log("info", "query " + requestId + " started");
+    log(
+      "info",
+      "query " + requestId + " started" + (request.partitionKeyRangeId ? " pkrid: " + request.partitionKeyRangeId : "")
+    );
     log("debug", request);
     const start = Date.now();
     const response = await executeRequest(request);
--- a/src/ClientContext.ts
+++ b/src/ClientContext.ts
@@ -1,7 +1,9 @@
+import uuid from "uuid";
 import { PartitionKeyRange } from "./client/Container/PartitionKeyRange";
 import { Resource } from "./client/Resource";
 import { Constants, HTTPMethod, OperationType, ResourceType } from "./common/constants";
 import { getIdFromLink, getPathFromLink, parseLink } from "./common/helper";
+import { logger } from "./common/logger";
 import { StatusCodes, SubStatusCodes } from "./common/statusCodes";
 import { CosmosClientOptions } from "./CosmosClientOptions";
 import { ConnectionPolicy, ConsistencyLevel, DatabaseAccount, PartitionKey } from "./documents";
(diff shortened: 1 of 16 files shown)


src/queryIterator.ts
```diff
@@ -217,4 +225,21 @@ export class QueryIterator<T> {
       error.substatus === SubStatusCodes.CrossPartitionQueryNotServable
     );
   }
+
+  private initPromise: Promise<void>;
```
southpolesteve: Discussed with you over teams. Lmk what you wanna do.

src/ClientContext.ts
```diff
@@ -1,7 +1,9 @@
+import uuid from "uuid";
```
southpolesteve: You want to import `uuid/v4` directly

src/request/FeedOptions.ts
```diff
@@ -12,17 +12,24 @@ export interface FeedOptions extends SharedOptions {
    * A user may set this field to allow for longer headers, which can help the backend optimize query execution."
    */
   continuationTokenLimitInKB?: number;
-  /** Allow scan on the queries which couldn't be served as indexing was opted out on the requested paths. */
+  /** Allow scan on the queries which couldn't be served as indexing was opted out on the requested paths. Default: false */
```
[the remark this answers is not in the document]
christopheranderson: People could be using IndexV1 with our new SDK if they are using a collection created with a legacy client.

src/queryIterator.ts
christopheranderson: Added new option to make this behavior change opt-in

PR #398, opened 2019-08-02
Fix bug where offset limit iterator was being called for each item under the offset count

southpolesteve:
Problem: Calling fetchNext() on a query iterator with an OFFSET/LIMIT query was returning "undefined" for each iteration up to the OFFSET count. This surfaced as a perf issue as the iterator was being called significantly more than needed in the portal to display results.

Solution: Properly aggregate results up to the OFFSET value and only return up the query pipeline when we have exhausted OFFSET


southpolesteve:
Added an options `expectedIteratorCalls` to cross partition query tests that will catch any regressions

PR #402, opened 2019-08-08
Add sample for bulk update with continuation token

southpolesteve:


samples/BulkUpdateWithSproc.ts
```diff
@@ -0,0 +1,87 @@
+import { logSampleHeader, handleError, finish, logStep } from "./Shared/handleError";
+import { CosmosClient } from "../dist";
+import { endpoint, key, database as databaseId, container as containerId } from "./Shared/config";
+import uuid from "uuid/v4";
+
+logSampleHeader("Bulk Update Using Stored Procedures");
+// Only to make TypeScript happy
+var getContext: any;
```
christopheranderson: [non-blocking] Should we be encouraging folks to write TS that gets compiled and then sent over the wire? I've experienced nothing but strife.

PR #405, opened 2019-08-13
Export default partition key path

southpolesteve:


christopheranderson:
Why do we need this exported?

southpolesteve:
We got asked for it in #308

